PR #35, opened 2014-05-11
Added jsonp support

arnold-almeida:
Heya. Here is a semi nice way of providing jsonp support without having to modify too much.

### DIFF
--- a/src/Exception/ResourceException.php
+++ b/src/Exception/ResourceException.php
@@ -8,14 +8,14 @@ class ResourceException extends HttpException {
 
 	/**
 	 * MessageBag errors.
-	 * 
+	 *
 	 * @var \Illuminate\Support\MessageBag
 	 */
 	protected $errors;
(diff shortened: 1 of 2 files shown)


j0an:
+1 @ingro

camilstaps:
@ingro @j0an 

All you need to do is:

1. Override Dingo's config.php (i.e. create a `/app/config/packages/dingo/api/config.php` to use the JsonP formatter:

        return [
           'default_format' => 'json',
            'formats' => [
                'json' => 'Dingo\Api\Http\ResponseFormat\JsonpResponseFormat',
            ]
        ];

2. Add `?callback=console.log` to the URL. this will call `console.log()` when the request is finished. 

Note that when `?callback` is not present, the formatter will default to normal json. Hope this helps others in the future, it took me an evening to figure this out. Perhaps this could be made more clear in the docs?

Issue #69, opened 2014-06-02
Allow disabling of automatic exception handling/add exception context


rizqidjamaluddin:
It can get tricky to debug problems if the exceptions getting thrown are silently getting transformed into their HTTP error/content counterparts. Maybe there can be a switch to disable this behavior, or use the app.debug config to attach the original exception's information along with the magic error if enabled?

PR #124, opened 2014-07-14
New API Debug Mode. 

hskrasek:
If enabled, any exceptions handled by Dingo Router will add an additional debug key filled with exception information.

This PR aims to solve the issue brought up by #69. There are somethings I am not sure about, such as the Config facade use, or the adding of HTTP Request information.

Up to discussing further to refine this. We could make this super basic, or even go as far as to write in custom formatting for the trace output. But don't want to overkill this.


olso:
Dead?

Issue #323, opened 2015-01-05
Autoload when using ?include on fractal collections


LeeMcNeil:
When using the ?include option on endpoints which return a fractal collection the dingo API is trying to autoload the requested relation even if it has not been placed in the availableIncludes array. Entering an include which does not exist causes an exception due to it trying to access a method on the model which doesn't exist.

Thanks :)

aikqzung:
+1
I think I'm facing this problem too.

jasonlewis:
Will #324 solve this?

LeeMcNeil:
As long as it checks if a method exists on the Eloquent model before attempting to call it, that PR should fix the issue.

MetalMatze:
I'm running into the same problem.

jasonlewis:
This should be fixed on `master` now.

Issue #146, opened 2014-08-07
Specifying scopes: ALL vs ANY


hipsterjazzbo:
Is it intentional that multiple scopes specified on a controller method results in ALL of the scopes being required, rather than ANY?

Issue #118, opened 2014-07-08
Eager loading on embeds


Anahkiasen:
Currently if you embed multiple relationships on a given request like `?foo,bar.baz` it will execute one query per entry. Wouldn't it be possible to check if the response is an instnace of `Illuminate\Database\Eloquent\Collection` and if yes to pass the includes to it directly like `$collection = $collection->load($this->includes)` ?

PR #324, opened 2015-01-05
Include changes

thomasvanlankveld:
A combination of $defaultIncludes and parsed $requestedIncludes is now pulled from the transformer for eager loading.

```php
protected $defaultIncludes = [
    'invite',
    'login',
];

protected $availableIncludes = [
    'posts',
];
```

Include keys can now be defined on the transformer to allow nested includes (A feature of Fractal that was no longer accessible in this package).

```php
protected $availableIncludes = [
    'group.admin' => 'admin',
];
```

Array syntax can now be used for requesting includes.
`users?include[]=login&include[]=invite`

*Update*
Nested includes using dot notation were [already available](https://github.com/thephpleague/fractal/issues/151#issuecomment-70772631) in Fractal, without you having to specify a key for it. They just don't work in `JsonApiSerializer` (For which I might submit a PR sometime in the future).

### DIFF
--- a/src/Transformer/FractalTransformer.php
+++ b/src/Transformer/FractalTransformer.php
@@ -158,9 +158,7 @@ class FractalTransformer implements TransformerInterface
      */
     public function getEagerLoads($transformer, $requestedIncludes)
     {
-        if(!is_array($requestedIncludes)) {
-            $requestedIncludes = array($requestedIncludes);
-        }
+        $requestedIncludes = (array) $requestedIncludes; // Ensure $requestedIncludeStrings is an array
 
         $availableRequestedIncludes = array_intersect($transformer->getAvailableIncludes(), $requestedIncludes);
         $defaultIncludes = $transformer->getDefaultIncludes();
--- a/src/Transformer/FractalTransformer.php
+++ b/src/Transformer/FractalTransformer.php
@@ -158,8 +158,9 @@ class FractalTransformer implements TransformerInterface
      */
     public function getEagerLoads($transformer, $requestedIncludes)
     {
-        if(!is_array($requestedIncludes))
+        if(!is_array($requestedIncludes)) {
             $requestedIncludes = array($requestedIncludes);
+        }
 
         $availableRequestedIncludes = array_intersect($transformer->getAvailableIncludes(), $requestedIncludes);
         $defaultIncludes = $transformer->getDefaultIncludes();
--- a/src/Transformer/FractalTransformer.php
+++ b/src/Transformer/FractalTransformer.php
@@ -150,10 +150,11 @@ class FractalTransformer implements TransformerInterface
     }
 
     /**
-     * Get includes as their array keys for eager loading
+     * Get includes as their array keys for eager loading.
      *
-     * @param  array|string $requestedIncludes
-     * @return array
+     * @param string|string[] $requestedIncludes
+     *
+     * @return string[]
      */
     public function getEagerLoads($transformer, $requestedIncludes)
     {
@@ -178,5 +179,4 @@ class FractalTransformer implements TransformerInterface
 
         return $eagerLoads;
     }
-
 }
--- a/src/Transformer/FractalTransformer.php
+++ b/src/Transformer/FractalTransformer.php
@@ -76,7 +76,9 @@ class FractalTransformer implements TransformerInterface
         }
 
         if ($response instanceof EloquentCollection) {
-            $response->load($this->fractal->getRequestedIncludes());
+            $requestedIncludes = $this->fractal->getRequestedIncludes();
+            $eagerLoads = $this->getEagerLoads($transformer, $requestedIncludes);
+            $response->load($eagerLoads);
         }
 
         foreach ($binding->getMeta() as $key => $value) {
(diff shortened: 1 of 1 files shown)


thomasvanlankveld:
Should `getEagerLoads()` be `protected`? I've opted for something like this to be included in Fractal [#146](https://github.com/thephpleague/fractal/issues/146) as well, so users could do something like:

```php
$users = User::with($eagerLoads)->get();
```

src/Transformer/FractalTransformer.php
```diff
@@ -129,7 +131,10 @@ protected function createResource($response, $transformer, array $parameters)
      */
     public function parseFractalIncludes(Request $request)
     {
-        $includes = array_filter(explode($this->includeSeparator, $request->get($this->includeKey)));
+        $includes = $request->get($this->includeKey);
+
+        if(!is_array($includes))
```
tymondesigns: shorthand conditionals not permitted in PSR-2

https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#51-if-elseif-else

jasonlewis:
This looks good but I want to pull it down locally and have a look at it more closely. Hopefully in the next couple of days or next week.

jasonlewis:
Can you see #323 and see if this PR will solve that as well. If not it shouldn't be too hard to implement a solution in this PR.

thomasvanlankveld:
@jasonlewis Cool. Checked and it solves [#323](https://github.com/dingo/api/issues/323) as well!

@bweston92 I used keys because this is how Fractal works. If you'd like a separate property you could suggest it [there](https://github.com/thephpleague/fractal).

thomasvanlankveld:
Could we also add a way to avoid this package's `->load()` altogether?

We're using `->load()`, which is actually 'lazy eager loading'. I notice that in some transformers I include foreign keys of related objects. This means these related objects are lazy loaded unless the API user requests their inclusion. This makes the requests very slow if the API user doesn't request all includes. I could solve this by eager loading my data before it enters the transformer (`User::with([])->get();`). But then, this package will still `->load()` them again later on. Since this makes the API slower, I'd like to be able to disable this package's `->load()`.

Maybe as an option argument to `->collection()`? Or as a config option?

```php
    /*
    |--------------------------------------------------------------------------
    | Eager Loading
    |--------------------------------------------------------------------------
    |
    | Included records can be eager loaded for better performance. Since
    | this package's implementation is actually "lazy eager loading", you may
    | want to disable it and implement true eager loading in your controller.
    |
    */

    'eager_loading' => true
```

jasonlewis:
@thomasvanlankveld What do you mean by avoid "this packages" `load`? Are you referring to the API package?

src/Transformer/FractalTransformer.php
```diff
@@ -76,7 +76,9 @@ public function transform($response, $transformer, Binding $binding, Request $re
         }
 
         if ($response instanceof EloquentCollection) {
-            $response->load($this->fractal->getRequestedIncludes());
+            $requestedIncludes = $this->fractal->getRequestedIncludes();
+            $eagerLoads = $this->getEagerLoads($transformer, $requestedIncludes);
+            $response->load($eagerLoads);
```
thomasvanlankveld: @jasonlewis This one right here.

I'm no performance expert though, I might be wrong about this.

stelgenhof:
When can we expect this to be included? Really would like to use the includes :)

src/Transformer/FractalTransformer.php
jasonlewis: Didn't even see this comment. My bad.

But that's the whole point of this. Is to eagerly load any requested includes. Otherwise they are lazily loaded. This was initially brought up in #118.

thomasvanlankveld: It gets a bit more complicated than that though. True eager loading changes `n + 1` into `1`, which is what Eloquent offers with `User::with('posts')->get();`. When you first do `User::all()` and then use `$users->load('posts)`, you are really executing `1 + 1` SQL queries, hence it's called ['lazy eager loading'](http://laravel.com/docs/4.2/eloquent#eager-loading). So maybe in cases where you really want just `1` query to get executed, you'd want to disable the `->load()`.

jasonlewis: True.

But it should only load in any additional relationships when they are specified by the includes key in the query string of the request, right?

thomasvanlankveld: Yup! (Or if they're specified in `$defaultIncludes` once this PR gets merged ;) )

jasonlewis: So it shouldn't be a problem... right?

thomasvanlankveld: No, now that I think of it, I can't imagine the query resulting from `->load()` ever to get that slow...

jasonlewis: Well, it shouldn't. Besides, I can't imagine why you wouldn't want to lazily eager load them. Saves running the loading yourself or loading them individually.

thomasvanlankveld: I got interested in true eager loading because in my API I'm pulling ids from related objects that may or may not be included, according to JSON-API spec.

In my `UserTransformer->transform()`:

```php
    if(count($user->posts)) {
        $links['posts'] = $user->posts->lists('id');
    }
```

This code is always called, so I get `n + 1` on the user index if I don't request the inclusion of posts.

jasonlewis: But that's not the issue here is it? That's just how it's going to be if you don't list `posts` as an include. Unless you're talking about something else that I'm not following on?

PR #330, opened 2015-01-09
Fix #322

### LINKED ISSUE #322: url('/') returns 'http://localhost' after using internal requests
When Im not on localhost, the `url('/')` should return the domain itself, which it does until usage of the [Internal Requests](https://github.com/dingo/api/wiki/Internal-Requests) like `$users = API::get('users/' . $id);`.

marktopper:
Fix #322

### DIFF
--- a/src/Facade/API.php
+++ b/src/Facade/API.php
@@ -85,4 +85,34 @@ class API extends Facade
     {
         return static::$app['api.response'];
     }
+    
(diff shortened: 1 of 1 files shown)


marktopper:
Might need some testing before approved

marktopper:
It seems to work just fine in all cases tried out.

PR #332, opened 2015-01-11
Allow item() and collection() to be called without a Transformer (IDE fix)

p4bloch:
When implementing the `TransformableInterface` in the model, there's no longer a need to pass a Transformer to the `item()` and `collection()` methods. This functionality is allowed by dingo, and no errors are thrown.

However, the IDE will squawk when I call `$this->response->item($model)` because the parameter doesn't have a default value set.
screen shot 2015-01-11 at 02 05 13

Defaulting $transformer to null fixed the IDE problem, and I can't think of a scenario where adding this will cause an error.

Let me know what you think @jasonlewis :)

### DIFF
--- a/src/Http/ResponseFactory.php
+++ b/src/Http/ResponseFactory.php
@@ -71,7 +71,7 @@ class ResponseFactory
      *
      * @return \Dingo\Api\Http\ResponseBuilder
      */
-    public function collection($collection, $transformer, array $parameters = [], Closure $after = null)
+    public function collection($collection, $transformer = null, array $parameters = [], Closure $after = null)
     {
         if ($collection->isEmpty()) {
             $class = get_class($collection);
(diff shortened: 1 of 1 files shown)


p4bloch:
Actually, when using `$this->response->collection($model)` without passing a Transformer, it *will* throw a missing argument Exception. After defaulting $transformer to null in `ResponseFactory`, the Transformer is loaded just fine from the one specified in the Model.

bweston92:
Pass the second argument as an empty model. The transformer gets passed a model instance therefore you won't be able to do $this->getAttribute, so I don't understand the real benefit but keeping this logic in the model. Single responsibility and all...

—

p4bloch:
I just wanted to point out that the call to `item()` and `collection()`, once you define the default Transformer in the Model, makes much more sense without any second argument at all. `$this->response->collection($collection)` looks way cleaner in my opinion :)

GrahamCampbell:
You need to update the docblock for this change. `object|null`.

jasonlewis:
`TransformableInterface` has been deprecated.

p4bloch:
Oh, I didn't notice that. I know updating the wiki must not be a top
priority task, but deprecated functionality should be updated as soon as
possible :) Is there any way we can help?

Thank you for your time!

Issue #336, opened 2015-01-13
CORS and strict mode


dbonner1987:
When "strict mode" is enabled, the router parses the Accept header on every request, unfortunately the OPTIONS request cannot be altered and as such, does not include a valid header. 

This immediately throws the "InvalidAcceptHeaderException" exception.

Suggest adding a check for the method in the parse method.

jasonlewis:
Why can't the OPTIONS request be altered?

dbonner1987:
Hi,

By design the preflight can't be altered, preventing the client from accessing a resource it shouldn't. The preflight OPTIONS request is constructed by the browser, all "author request header" values are ignored and the type is set added to the value of the Access-Control-Request-Headers (http://www.w3.org/TR/cors/#preflight-request).

So because the Accept value is ignored on preflight, when strict mode is enabled the request throws a 500.

For example, take a simple request like, not taking authentication into account:

> Request URL:http://api.domain.dev/products
> Request Method:GET
> Accept:application/vnd.vendorname.v1+json
> Accept-Encoding:gzip, deflate
> Accept-Language:en-US,en;q=0.8
> Connection:keep-alive
> Content-Length:46
> Content-Type:application/json;charset=UTF-8
> Host:api.domain.dev
> Origin:http://app.domain-2.dev
> Referer:http://app.domain-2.dev/

The preflight the browser will send is:

> Request URL:http://api.domain.dev/products
> Request Method:OPTIONS
> Accept:\*/\*
> Accept-Encoding:gzip, deflate, sdch
> Accept-Language:en-US,en;q=0.8
> Access-Control-Request-Headers:accept
> Access-Control-Request-Method:GET
> Connection:keep-alive
> Host:api.domain.dev
> Origin:http://app.domain-2.dev
> Referer:http://app.domain-2.dev/

Github has a similar example: https://developer.github.com/v3/#cross-origin-resource-sharing

The OPTIONS request is purely to establish what capabilities/permissions are valid for CORS interaction, altering the "parseAcceptHeader" method seems like the simplest fix. The vendor/version isn't required for such a request.

Cheers mate

dbonner1987:
I've just pushed a simpler solution that checks for the OPTIONS request at the start of the dispatch method whilst checking if it is an api request. 

Laravel has built-in support to handle these request automatically, this prevents the package having to run through and build a response of it's own.

Cheers

jasonlewis:
[...]

PR #337, opened 2015-01-13
Updates Accept Header parsing

dbonner1987:
Fixes the issue with strict mode and CORS, allowing the pre-flight OPTIONS request to be handled appropriately. #336

### DIFF
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -438,7 +438,7 @@ class Router extends IlluminateRouter
     {
         if (preg_match('#application/vnd\.'.$this->config->getVendor().'.(v[\d\.]+)\+(\w+)#', $request->header('accept'), $matches)) {
             return array_slice($matches, 1);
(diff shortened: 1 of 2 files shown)


src/Routing/Router.php
```diff
@@ -216,7 +216,7 @@ public function addExistingRoutes(IlluminateRouteCollection $routes)
      */
     public function dispatch(Request $request, $morph = true)
     {
-        if (! $this->isApiRequest($request)) {
+        if (! $this->isApiRequest($request) OR $request->method() == 'OPTIONS') {
```
lucasmichot: `||` instead or `OR` would be sufficient

dbonner1987: @lucasmichot true. An artefact of an earlier idea, where I was handing the 'OPTIONS' response elsewhere based on the api request and not just handing any 'OPTIONS' request off to Laravel. Branch has been updated.

Issue #321, opened 2015-01-05
Scopes in router should be a array ?!


JasonXt:
Route::api(['protected'=>true,'scopes'=>'student'],function(){});  This comes with a error  in LeagueOAuth2Provider.php Line 125  , the $scopes is just a string ,can not be foreach

when changed to below is ok!

Route::api(['protected'=>true,'scopes'=>['student']],function(){});

Is this  a bug?

jasonlewis:
If the docs mention you can pass it in as a string then it should be able to accept the string. So it's most likely a bug. Feel free to send a PR if you can.

jasonlewis:
Pushed a fix for this.

PR #340, opened 2015-01-17
route scope accept string #321

yohang88:
bug fix for accepting string in scope route


lucasmichot:
tests ?

jasonlewis:
I've pushed a different fix for this that fixes it at the core instead of in the individual providers. Cheers.

PR #355, opened 2015-02-02
fixed accept header regex

gxela:


jasonlewis:
How are you doing your versions?

gxela:
I would like to be able to have this flexibility in my api versioning

gxela:
It doesn't hurt to add this, It allows me to add more detailed versioning to my architecture and so I would like to not have to modify this file everytime I composer update

bweston92:
Create your own service provider for dingo and construct your own router that extends dingo's router.



—

jasonlewis:
For the L5/Lumen versions you can overload the Accept parser now and do it how you want.

PR #364, opened 2015-02-17
camel casing on exception output for consistency

specialtactics:
For the sake of API niceless and consistency, changes such that when exceptions are thrown, the error data has camel cased keys.

### DIFF
--- a/src/Event/ExceptionHandler.php
+++ b/src/Event/ExceptionHandler.php
@@ -61,10 +61,18 @@ class ExceptionHandler
             $message = sprintf('%d %s', $exception->getStatusCode(), Response::$statusTexts[$exception->getStatusCode()]);
         }
 
-        $response = ['message' => $message, 'status_code' => $exception->getStatusCode()];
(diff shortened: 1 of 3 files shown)


jasonlewis:
What's consistent for you may not be consistent for others. I'm open to a discussion on this but I think the best approach would be to make it easily configurable in some way.

specialtactics:
Sure, it wouldn't be hard to have a switch in the config for camel casing exception output.

The reason I feel it's consistent is because JsonReponseFormat does the same thing in relation to array keys, and the package should provide an overall consistent convention so far as camel case vs snake case of the output keys.

jasonlewis:
I think the response formatter will adjust the keys based on models only and whether or not snake casing the attributes is enabled/disabled. I'll look at some way of making it configurable as some people probably want more control over how the errors are returned without replacing the entire exception handler.

jasonlewis:
In L5/Lumen versions you can set the default format for a generic error response.

PR #370, opened 2015-03-04
Ensure Code style is PR2

lucasmichot:


GrahamCampbell:
90% of these changes are NOT PSR-2 specified.

GrahamCampbell:
Actually, it's 99% of this that has nothing to do with PSR-2 at all! Only the change to the namespace to match the folder name is PSR-2. Everything else is not PSR-2.

PR #372, opened 2015-03-04
updating jwt-auth and oauth2

tymondesigns:
Thought I would get the ball rolling a little bit here :)

I have updated jwt-auth to use `0.5.*` and updated the tests.
Also updated lucadegasperi/oauth2-server-laravel to `4.0.x@dev` for now.

Obviously tests are failing at this point (Auth ones pass), but don't error anymore :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,8 +15,8 @@
         "league/fractal": "0.11.*"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


jasonlewis:
The L5 branch is on hold for a little bit. Things are a bit more complicated than I initially thought they would be. Cheers.

PR #374, opened 2015-03-05
Ensure that API filters are the first executed

v-six:
Hi,

I encountered an issue with Laravel route group filters and dingo API basic authentification provider today.

Let's suppose that we are requesting `/api/foo/bar` with a valid HTTP basic authentification header and that we have this `routes.php` file :
```php
Route::api(
    [
        'version' => 'v1',
        'prefix' => 'api',
        'protected' => true,
    ],
    function () {
        Route::group(
        [
            'prefix' => 'foo',
            'before' => 'foo',
        ],
        function () {
             Route::get(
                 '/bar',
                 [
                     'as' => 'foo.bar',
                     'uses' => 'FooController@bar'
                 ]
             );
        }
    }
);
```

In that case, I observe that the `foo` filter is executed before the user basic authentification one. But with this behavior, I can't do any user validation in `foo` filter because no user is logged in yet.

I think that every API filters should be executed before any others to avoid this issue.

Laravel has this routing behavior (in a short form ^^) :
```php
$action = Route::parseAction($routeData);
$groupAction = $groupData + $action;
$fullAction = Route::parseAction($groupAction);
```

So the API filters must be deleted (if exist) and placed in the beginning of filters array in every `Route::parseAction()` call.

### DIFF
--- a/src/Routing/Route.php
+++ b/src/Routing/Route.php
@@ -42,9 +42,12 @@ class Route extends IlluminateRoute
         }
 
         foreach ([static::API_FILTER_THROTTLE, static::API_FILTER_AUTH] as $filter) {
-            if (array_search($filter, $action['before']) === false) {
-                array_unshift($action['before'], $filter);
+            $key = array_search($filter, $action['before']);
+            if ($key !== false) {
+                unset($action['before'][$key]);
             }
+
+            array_unshift($action['before'], $filter);
         }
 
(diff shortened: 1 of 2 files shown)


jasonlewis:
Cheers mate, sorry it took so long to merge!

kevindierkx:
This should resolve #387, nice one!

PR #376, opened 2015-03-07
made dispatcher tests pass

willishq:
fixed some small errors which were makign the dispatcher tests fail


willishq:
obviously there are more than just the dispatcher tests failing but I managed to get them passing and the dispatcher appears to be working for me too.

jasonlewis:
Not merging anything L5 related as yet as the upgrade is going to be a little more complicated than I initially thought it would be. Cheers.

Issue #369, opened 2015-03-03
JWT Authentication error with new version


gdg87:
Hi, I update [jwt-auth](https://github.com/tymondesigns/jwt-auth) to the last release (0.4.1) but the authentication filter doesn't work more. 
I see that they changed the `login()` method with `authenticate()`. 
Change JWTProvider on line 46

    return $this->auth->login($token);
   
  with 

      return $this->auth->authenticate($token);
  works fine.

tymondesigns:
yea dingo currently requires version `0.3.*` i'm afraid. So you won't be able to use the latest jwt-auth with it.

I'm not sure whether @jasonlewis is prepared to release a new version just for that.

jasonlewis:
Aye, the L5 version which I'll get pushed up soon will be the latest JWT
authentication.

Andreyco:
No fix for this? Somehow... I cannot upgrade to L5.

jasonlewis:
This is fixed in `0.8.*@dev`, tagging release soon. No news on L5 for now.

PR #377, opened 2015-03-08
An update for JWTProvider.php

numnine:
Login() renamed to authenticate() to support jwt 0.4.* and above


tymondesigns:
See #369

jasonlewis:
Fixed, will tag soon.

PR #386, opened 2015-03-28
Update the authenticate function in the Dingo\Api\Auth\JWTProvider 

sunel:
Fixed the authenticate function according to "tymon/jwt-auth": "0.4.*" library


tymondesigns:
See #377 and #369

jasonlewis:
I've now fixed this. Will tag soon.

PR #393, opened 2015-04-22
added responseFormatInterface

codbox:
In Response class \Dingo\Api\Http\Format\FormatInterface was type hinted but the interface is not present. 

hence created \Dingo\Api\Http\ResponseFormat\ResponseFormatInterface class. 
```
  /**
     * Get the formatter based on the requested format type.
     *
     * @param string $format
     *
     * @throws \RuntimeException
     *
     * @return \Dingo\Api\Http\Format\FormatInterface
     */
    public static function getFormatter($format)
    {
        if (! static::hasFormatter($format)) {
            throw new NotAcceptableHttpException('Unable to format response according to Accept header.');
        }

        return static::$formatters[$format];
    }
```

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -104,7 +104,7 @@ class Response extends IlluminateResponse
      *
      * @throws \RuntimeException
      *
-     * @return \Dingo\Api\Http\Format\FormatInterface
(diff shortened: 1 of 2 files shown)


jasonlewis:
This would've been an error in the docblock. Shifting my focus to L5 and Lumen so things here have changed.

PR #400, opened 2015-05-17
Add Laravel 5.1 support

vinkla:
Since Laravel 5.1 is just around the corner I think it would be good if we already started to support it. Shouldn't break anything.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,12 +11,12 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "illuminate/routing": "5.0.*",
-        "illuminate/support": "5.0.*",
+        "illuminate/routing": "5.0.*|5.1.*",
+        "illuminate/support": "5.0.*|5.1.*",
         "league/fractal": "0.12.*"
     },
     "require-dev": {
-        "lucadegasperi/oauth2-server-laravel": "4.0.*@dev",
+        "lucadegasperi/oauth2-server-laravel": "4.1.*@dev",
         "tymon/jwt-auth": "0.5.*",
         "illuminate/auth": "5.0.*",
         "illuminate/cache": "5.0.*",


composer.json
```diff
@@ -11,19 +11,19 @@
...
-        "lucadegasperi/oauth2-server-laravel": "4.0.*@dev",
+        "lucadegasperi/oauth2-server-laravel": "4.1.*@dev",
         "tymon/jwt-auth": "0.5.*",
-        "illuminate/auth": "5.0.*",
-        "illuminate/cache": "5.0.*",
-        "illuminate/database": "5.0.*",
-        "illuminate/pagination": "5.0.*",
-        "illuminate/console": "5.0.*",
-        "illuminate/filesystem": "5.0.*",
+        "illuminate/auth": "5.0.*|5.1.*",
```
GrahamCampbell: This change is wrong.

GrahamCampbell: For require-dev, pick the version you want to use, don't allow both.

vinkla: Okay, I'll update.

vinkla:
Maybe we should skip `5.0` support all together? Since Laravel 5.1 is just around the corner.

GrahamCampbell:
:+1: for skipping 5.0 and only supporting 5.1.

AndyWendt:
:+1:  for skipping 5.0

I doubt many will stay on 5.0 for any length of time

GrahamCampbell:
5.1 will be LTS too. :)

AndyWendt:
Exactly :smile: hopefully we will not have a panic like we did with 5.1

jasonlewis:
I might just aim for 5.1 in that case then I think. I doubt there's much that's changed that'd affect this package anyway.

PR #402, opened 2015-05-17
Test on PHP 7.0

vinkla:
Still allowing failures though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,13 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7.0
   - hhvm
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:-1: on this given that mockery doesn't work on php 7 properly yet.

vinkla:
@GrahamCampbell Does it matter? Since we still allow it to fail.

GrahamCampbell:
There's no point in adding something that fatal errors out of our control. Not doing any good at all.

vinkla:
Closing this due to lack of interest.

PR #403, opened 2015-05-17
Update deprecated PHPUnit configuration and rename XML

vinkla:
Update strict behavior and renamed `phpunit.xml` to `phpunit.xml.dist`.


GrahamCampbell:
:+1:

PR #404, opened 2015-05-17
Move setup config to boot method

vinkla:
Also added publish method to move the config file to /config folder in a Laravel project.

### DIFF
--- a/src/Provider/ApiServiceProvider.php
+++ b/src/Provider/ApiServiceProvider.php
@@ -13,13 +13,22 @@ use Dingo\Api\Transformer\Factory as TransformerFactory;
 class ApiServiceProvider extends ServiceProvider
 {
     /**
+     * Boot the service provider.
+     *
+     * @return void
+     */
+    public function boot()
+    {
+        $this->setupConfig();
+    }
+
+    /**
      * Register the service provider.
      *
      * @return void
      */
     public function register()
     {
-        $this->setupConfig();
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:+1: It's important this is done AFTER the user has a chance to override it.

jasonlewis:
The publishing is done in the Laravel service provider as it will error it for Lumen installs. So you can remove that. `vendor:publish` should already work.

GrahamCampbell:
This isn;t just for publishing.

src/Provider/ApiServiceProvider.php
```diff
@@ -42,7 +51,11 @@ public function register()
      */
     protected function setupConfig()
     {
-        $this->mergeConfigFrom(__DIR__.'/../../config/api.php', 'api');
+        $source = realpath(__DIR__.'/../../config/api.php');
+
+        $this->publishes([$source => config_path('api.php')]);
```
jasonlewis: This looks an awful lot like publishing to me @GrahamCampbell.

It's already done here: https://github.com/dingo/api/blob/develop/src/Provider/LaravelServiceProvider.php#L19

jasonlewis:
The `config_path` function isn't available in Lumen. That's why publishing isn't in this provider. I'm fine with the other changes.

jasonlewis:
I'll fix it up on my end. Cheers!

PR #415, opened 2015-05-20
Fixing paginator class in response factory

stygiansabyss:
This seems to accomplish the goal.  Switches the class to using the Paginator class that fractal expects and snags the collection in the Response to properly display the results.

### DIFF
--- a/src/Http/Response/Factory.php
+++ b/src/Http/Response/Factory.php
@@ -7,7 +7,7 @@ use ErrorException;
 use Illuminate\Support\Str;
 use Dingo\Api\Http\Response;
 use Illuminate\Support\Collection;
-use Illuminate\Pagination\Paginator;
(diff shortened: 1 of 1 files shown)


src/Http/Response/Factory.php
```diff
@@ -124,9 +124,10 @@ public function paginator(Paginator $paginator, $transformer, array $parameters
 
         $binding = $this->transformer->register($class, $transformer, $parameters, $after);
 
-        return new Response($paginator, 200, [], $binding);
+        return new Response($paginator->getCollection(), 200, [], $binding);
     }
 
+
```
GrahamCampbell: ?

stygiansabyss:
Pretty sure this whole PR wont work quite like I expected.  The result has the proper amount of results, but does not contain the fractal metadata for pagination.

Issue #419, opened 2015-05-25
Additional fields in error responses. Exceptions with the any additional data.


KIVagant:
**Problem:**

I need to add some special fields to error request. The array 'errors' in Dingo custom Exceptions is not enough for me.

For example, i need to send error message:
```
"message": "Oops, some fields looks like incorrect!",
"system_message": "Validation error",
"something": "else",
"yet": ["another", "field"],
"code": 1234,
"status_code": 422,
```

**Resolution:**

In api.php setup new format:
```
'errorFormat' => [
        'message' => ':message',
        'system_message' => ':system_message', // this is a new one
        'errors' => ':errors', // this is an old, but can be used not only in ResourceException childs
...
    ],
```

Create the new Exception:

```
class TestException extends HttpException implements DingoException
{
    /**
     * @var \Illuminate\Support\MessageBag
     */
    protected $errors;
    protected $system_message = '';
    public function __construct($message = null, $system_message = '', $errors = null, Exception $previous = null, $headers = [], $error_code = 0)
    {
        $this->system_message = $system_message;
        if (is_null($errors)) {
            $this->errors = new MessageBag;
        } else {
            $this->errors = is_array($errors) ? new MessageBag($errors) : $errors;
        }

        parent::__construct(400, $message, $previous, $headers, $error_code);
    }

    // This is a new desired functionality
    public function getReplacements()
    {
        return [
            ':errors' => $this->errors,
            ':system_message' => $this->system_message,
        ];
    }
}
```

Than throw it from API

```
throw new TestException('Lovely bug', 'All is dead', [ 'humans' => 'dead', 'cockroaches' => 'dead' ], null, [], 123456);
```

And now, get this result:
```
{
    "message": "Lovely bug",
    "system_message": "All is dead",
    "errors": {
        "humans": [
            "dead"
        ],
        "cockroaches": [
            "dead"
        ]
    },
    "code": 123456,
    "status_code": 400
}
```

Pull request is coming.

jasonlewis:
Just catch all errors thrown and deal with them yourself. You can use
app('api.exception')->register() with a call back that type hints the
global Exception class.

KIVagant:
@jasonlewis , if I will catch exception before dingo, then I will need to rewrite all this logic:

```
 return $this->prepareResponse(
                $this->exception->handle($exception),
                $request,
                $accept['format']
            );
```

It is not good way.

[...]

PR #420, opened 2015-05-25
Exceptions now can return any additional data

KIVagant:
#419 Exceptions which implements DingoException will allow to setup additional replacements.

**Problem:**

I need to add some special fields to error request. The array 'errors' in Dingo custom Exceptions is not enough for me.

For example, i need to send error message:
```
"message": "Oops, some fields looks like incorrect!",
"system_message": "Validation error",
"something": "else",
"yet": ["another", "field"],
"code": 1234,
"status_code": 422,
```

**Resolution:**

In api.php setup new format:
```
'errorFormat' => [
        'message' => ':message',
        'system_message' => ':system_message', // this is a new one
        'errors' => ':errors', // this is an old, but can be used not only in ResourceException childs
...
    ],
```
Use default Exceptions *OR* create the new Exception with public property $replacements;

```
$throw = new \Symfony\Component\HttpKernel\Exception\BadRequestHttpException('Lovely bug', null, 123456);

// Here is a new functionality
$throw->replacements = [
    ':errors' => [ 'humans' => 'dead', 'cockroaches' => 'dead' ],
    ':system_message' => $error['system_message'],
    // here you can add any fields, but don't forget to setup them in api.php
];
throw $throw;
```

And now, get this result:

```
{
    "message": "Lovely bug",
    "system_message": "All is dead",
    "errors": {
        "humans": [
            "dead"
        ],
        "cockroaches": [
            "dead"
        ]
    },
    "code": 123456,
    "status_code": 400
}
```

### DIFF
--- a/src/Exception/ExceptionReplacementsTrait.php
+++ b/src/Exception/ExceptionReplacementsTrait.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Dingo\Api\Exception;
+
+/**
+ * Class ExceptionReplacementsTrait
+ *
+ * Example:
+ * <code><?php
+ * class MyException extends HttpException
+ * {
+ *      use ExceptionReplacementsTrait;
+ * //...
+ * }
+ * $ex = new MyException(); // somewhere in API
+ * $ex->setReplacements([':error' => $errors]);
+ * throw $ex;
+ * ?>
+ * @package Dingo\Api\Exception
+ */
+class ExceptionReplacementsTrait {
+
+    /**
+     * Array of replacements from config/api.php and their values
+     * @var array
+     */
+    public $replacements = [];
+
+    /**
+     * @return array
+     */
+    public function getReplacements()
+    {
+        return $this->replacements;
+    }
+
+    /**
+     * @param array $replacements
+     */
+    public function setReplacements($replacements)
+    {
+        $this->replacements = $replacements;
+    }
+}
--- a/src/Exception/Handler.php
+++ b/src/Exception/Handler.php
@@ -167,7 +167,9 @@ class Handler implements ExceptionHandler
             ':message' => $message,
             ':status_code' => $statusCode
         ];
-
+        if (isset($exception->replacements)) {
+            $replacements += $exception->replacements;
+        }
         if ($exception instanceof ResourceException && $exception->hasErrors()) {
             $replacements[':errors'] = $exception->getErrors();
         }
--- a/src/Exception/Handler.php
+++ b/src/Exception/Handler.php
@@ -167,7 +167,9 @@ class Handler implements ExceptionHandler
             ':message' => $message,
             ':status_code' => $statusCode
         ];
-        if (isset($exception->replacements)) {
+        if (method_exists($exception, 'getReplacements')) {
+            $replacements += $exception->getReplacements();
+        } elseif (isset($exception->replacements)) {
             $replacements += $exception->replacements;
         }
         if ($exception instanceof ResourceException && $exception->hasErrors()) {


KIVagant:
Changed:
- Interface DingoException was removed;
- Public property 'replacements' is used instead of method.

jasonlewis:
I do like this idea but I'm not so sure on the implementation. I think going the route of an interface is a nicer and more obvious approach. The `ResourceException` can be adjusted to implement the interface by default and the end user is able to create their own exceptions as well.

This as it stands doesn't feel right.

So my proposal would be to create a new interface, which you already had by the sounds of it, at `Dingo\Api\Exception\ErrorReplacements`. Like you had earlier, it should require a `getReplacements` and also a `setReplacements`  method.

The `ResourceException` can be updated to implement this interface and provide the default methods.

How's that sound? (I have a feeling this is pretty much how you had it to begin with.)

KIVagant:
I'm understand your desire for the ideal solution, but answer me to one question, please:
How I can throw a _Symphony HTTPException_ with your way?

Look at [ this method](https://github.com/mitulgolakiya/laravel-api-generator/pull/54/files#diff-d241c965bad2143a2a565eb7174157b7R131)

In the current implementation of the PR, I do not force the user to extend all exceptions only for pass an additional parameter in the server's response. 

Even if I setup the method like this,:
```
$ex = new BadRequestHttpException('some_message');
$ex->getReplacements = function (){return $replacements};
```
..., you will can't check it inside Handler::genericResponse()
PHP don't allow to call method like $ex->getReplacements(), and method_exists will not show it, and BadRequestHttpException does not implement some special Interface for instance checking.

So there we will only have one way to implement basic problem: make many exception classes with traits or with implementing interfaces. But why library user really need to know than we are make any Exceptions or requiring the user to create them every time when he will need to send some ID to response?

You already have excess class ResourceException and [here I said](https://github.com/dingo/api/issues/419#issuecomment-105263452), how you can make it more simplest and readable. Really, in laravel-api-generator example I don't need to use it anymore, because i can throw _any information that I need_ just with little wrapper helpers and _default Symphony Exceptions_.

Just a use case: I can [easy move this](https://github.com/mitulgolakiya/laravel-api-generator/pull/54/files#diff-d241c965bad2143a2a565eb7174157b7R14) (and I will do) to config and user will don't know, that we make a little magic inside. He just can setup any HttpException inside 'exception' config in error register.

If you want a perfect variant with OOP classic, then you need, at least, rewrite Handler::genericResponse(). I proposed short variant without change many code.

My basic message: the library should resolve their goals, but don't make a limitations or yet another code conventions. We can add a special trait with setReplacements() method inside, but just for coders, who want a perfectionism.

KIVagant:
https://github.com/dingo/api/commit/6f946d01a88c8191ec4ee0967676749b396b77c9

jasonlewis:
I'm still not sold on how this is implemented. I think I can see where you're coming from in terms of it being a pain to extend all exceptions for only a little thing. I'm going to play with some things and see if I can come up with an extendible solution that doesn't involve setting an arbitrary public method or using an interface.

jasonlewis:
How about this, in an effort to keep the core of this package "clean", you can set a custom replacements callback on the handler.

The handler can then fire the callback and merge your custom replacements with the others. Example...

```php
app('api.exception')->setCustomReplacements(function ($exception) {
    return $exception->replacements;
});
```

The callback must simply return an array of key/value pairs of replacements.

KIVagant:
Maybe just 
```
app('api.exception')->setCustomReplacements($replacements);
throw new SomeHttpException();
``` 
?

jasonlewis:
Yeah sure, that way you don't even need to set the public property I suppose. Win win.

KIVagant:
Ok, I like this variant and looking forward for implementation!

jasonlewis:
I've just pushed the implementation. Method is `setReplacements` on the exception handler instance, available through `app('api.exception')`.

PR #429, opened 2015-05-31
Add a PHP-CS-Fixer definition file

lucasmichot:
Anyone can run `php-cs-fixer fix` before commiting

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,9 @@
+<?php
+
+$finder = Symfony\CS\Finder\DefaultFinder::create()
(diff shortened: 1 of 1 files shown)


jasonlewis:
Running the fixer updates all the test files and replaces the namespace with a lowercase `tests`. I think I understand why it's doing it but I've got the autoload setup in the `composer.json` file pointing directly to the directory to it doesn't make sense. Anyway we can fix that?

vinkla:
@jasonlewis I've pushed a fix for this in #434

PR #434, opened 2015-06-01
Fix PHP CS Fixer namespace bug

vinkla:
This fixes a bug introduced in #429.


GrahamCampbell:
The correct solution is to disable the psr0 fixer since we're not using psr0...

vinkla:
@GrahamCampbell Isn't it included by default with the PSR2 level? How do we disable it?

GrahamCampbell:
> How do we disable it?

Just put `-psr0` into the fixers list.

PR #435, opened 2015-06-01
Add PHP CS Fixer to composer file

vinkla:
Add the package to composer.json as introduced in #429

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -5,5 +5,5 @@ $finder = Symfony\CS\Finder\DefaultFinder::create()
 
 return Symfony\CS\Config\Config::create()
     ->level(Symfony\CS\FixerInterface::PSR2_LEVEL)
-    ->fixers(array('short_array_syntax'))
+    ->fixers(['short_array_syntax'])
     ->finder($finder);
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
         "illuminate/filesystem": "5.0.*",
         "phpunit/phpunit": "~4.0",
         "mockery/mockery": "~0.9",
-        "squizlabs/php_codesniffer": "~2.0"
+        "squizlabs/php_codesniffer": "~2.0",
+        "fabpot/php-cs-fixer": "1.9.*@dev"
     },
     "suggest": {
         "lucadegasperi/oauth2-server-laravel": "Protect your API with OAuth 2.0.",


composer.json
```diff
@@ -26,7 +26,8 @@
         "illuminate/filesystem": "5.0.*",
         "phpunit/phpunit": "~4.0",
         "mockery/mockery": "~0.9",
-        "squizlabs/php_codesniffer": "~2.0"
+        "squizlabs/php_codesniffer": "~2.0",
+        "fabpot/php-cs-fixer": "1.9.*"
```
lucasmichot: No PHP-CS-Fixer should be installed globally.

composer.json
```diff
@@ -46,5 +47,7 @@
         "branch-alias": {
             "dev-develop": "0.9-dev"
         }
-    }
+    },
+    "minimum-stability": "dev",
```
lucasmichot: Not part of this feature

.php_cs
```diff
@@ -5,5 +5,5 @@ $finder = Symfony\CS\Finder\DefaultFinder::create()
 
 return Symfony\CS\Config\Config::create()
     ->level(Symfony\CS\FixerInterface::PSR2_LEVEL)
-    ->fixers(array('short_array_syntax'))
+    ->fixers(['short_array_syntax'])
```
lucasmichot: No. PHP-CS-Fixer is still base on PHP >= 5.3.6
This should be long array syntax
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.9/composer.json#L17

composer.json
vinkla: It is if we want to use `1.9.*`.

composer.json
GrahamCampbell: Yeh, I agree.

composer.json
GrahamCampbell: No, we can just change the version constraint above to `1.9.*@dev`.

composer.json
GrahamCampbell: Why not just use https://styleci.io/ and ditch the cs checks from the tests. The tests and the cs should be separate. StyleCI will still use github's status API on pull requests.

composer.json
vinkla: Okay, I'll fix.

.php_cs
GrahamCampbell: @lucasmichot So? We're only running php-cs-fixer on php 5.4+ in this context.

composer.json
jasonlewis: Easy to add in @GrahamCampbell?

vinkla: @jasonlewis Just sign up at https://styleci.io/ and add the repository.

GrahamCampbell: > Easy to add in @GrahamCampbell?

Yeh, it is, it at least I hope it is. Ping me if you need any help. :)

PR #437, opened 2015-06-01
Add GIT attributes file

vinkla:
In production we don't want to download the test files.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,8 @@
+/tests export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.travis.yml export-ignore
+/phpunit.xml.dist export-ignore
+/CHANGELOG.md export-ignore
+/CONTRIBUTING.md export-ignore
+/README.md export-ignore


lucasmichot:
Probably lots of other files at the root to add to `.gitignore`

vinkla:
Yeah, we could add everything except the `composer.json`, `config` and `src` if we want to save more space. @jasonlewis What do you think?

lucasmichot:
I would add .editorconfig, .travis.yml and phpunit.xml

vinkla:
I would go with either just the test folder or something like this;

```
/tests export-ignore
/.gitattributes export-ignore
/.gitignore export-ignore
/.travis.yml export-ignore
/phpunit.xml.dist export-ignore
/CHANGELOG.md export-ignore
/CONTRIBUTING.md export-ignore
/README.md export-ignore
```

jasonlewis:
Yeah all the cruff can probably be ignored.

vinkla:
Done.

PR #439, opened 2015-06-01
Add license file

vinkla:


LICENSE
```diff
@@ -0,0 +1,27 @@
+Copyright (c) 2015, Jason Lewis
```
GrahamCampbell: The year should be as low as possible.

vinkla: Should we maybe do a 2014-2015 instead?

LICENSE
```diff
@@ -0,0 +1,27 @@
...
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+
+* Neither the name of tewtr1we nor the names of its
```
jasonlewis: `tewtr1we`?

LICENSE
jasonlewis: Either should be fine.

LICENSE
vinkla: Fixed.

PR #441, opened 2015-06-01
Added styleci config

GrahamCampbell:


vinkla:
@GrahamCampbell Should we remove the `.php_cs` as well?

GrahamCampbell:
Nah, that's still useful for people running php-cs-fixer locally.

vinkla:
Ok.

PR #448, opened 2015-06-03
Fix Paginator reference to LengthAwarePaginator

tymondesigns:
Cannot get the paginator response to work using the response factory because fractal requires an instance of `LengthAwarePaginator` in the `IlluminatePaginatorAdapter` [here](https://github.com/thephpleague/fractal/blob/master/src/Pagination/IlluminatePaginatorAdapter.php#L38).

All works great after this change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,3 @@
-/vendor
-/coverage
-/.idea
-phpunit.xml
-composer.phar
 composer.lock
-.DS_Store
+phpunit.xml
+vendor


jasonlewis:
Is it perhaps worth type-hinting the `AbstractPaginator` instead of `LengthAwarePaginator` in case people aren't using Fractal but would like to use the regular `Paginator` instance?

tymondesigns:
Yep, that would be better I think.. Will tweak and rebase when I'm home

jasonlewis:
All good mate. Cheers.

tymondesigns:
Messed up my history 😪 Tried to target master as everything has been merged over..

#454 now targets master and has the changes as suggested

Issue #398, opened 2015-05-17
Lumen AND Laravel 5 Support


jasonlewis:
Hello all, I've recently pushed quite a number of changes to the `develop` branch that allows you to use this package in your Lumen and Laravel 5 apps.

You can require it using `0.9.*@dev`.

Please note that if you're using Lumen you'll need to be using `lumen-framework 5.0.*@dev` for it to work correctly. These several routing changes that are yet to make it to a stable Lumen release.

A few things to note...

1. Much of the configuration can be done via environment variables now. You can still publish the config file in Laravel 5 and make changes that way. Lumen will require you to use the environment variables. For things that can't be set with environment variables there will be a method available on the underlying instance to change whatever it is you need.
2. There is no internal dispatcher at this point in time. I was focusing on the core of the API for now to make that that's all sweet before moving on to the internal dispatcher.
3. There is no route command at this point in time. It will be back.
4. There is no URL generator at this point in time. it will be back.
5. There's no facade at the moment. You will need to get an instance of the API router (`app('api.router')`) to register your routes.

Here is a very quick sample of how it's now used. In your Lumen `bootstrap/app.php` file you'd register the Lumen provider.

```php
$app->register('Dingo\Api\Provider\LumenServiceProvider');

// Get an instance of the API router so we can use it in the routes.php file
$api = $app['api.router'];
```

Now we can register our routes in the `routes.php` file.

```php
$api->version('v1', function ($api) {
    $api->resource('users', 'UserController');
});
```

All routing methods available in L5 are available in both L5 and Lumen. This includes standard routing with `get`, `post`, etc, as well as `resource` and `controller` routes.

You can also setup group options much like in previous options (there is also an `$api->group()` method).

```php
$api->version('v1', ['namespace' => 'App', 'protected' => true], function ($api) {
    $api->resource('users', 'UserController');
});
```

Aside from that not a whole lot has changed but there are a few new things that I'll be talking about over time. For now, I'd love it if people could give it a whirl and let me know if they come across any issues. I'll iron things out as quick as I can.

Cheers.

---

Another thing to note is that a `prefix` or `domain` will be required now. You can set these via the environment variables `API_PREFIX` or `API_DOMAIN` respectively, or by publishing the config and making the appropriate changes.

develpr:
<3 Thank you Jason!

jinqibu:
thanks a lot!

devmark:
thanks!

bweston92:
[...]

PR #452, opened 2015-06-04
Merge config on register

antonioribeiro:
As far as I can tell, you have to merge your config on register() and not on boot().

Before fixing it, on a clean install, I was also getting a

```
ErrorException in Handler.php line 53:
Argument 2 passed to Dingo\Api\Exception\Handler::__construct() must be of the type array, null given, called in /home/vagrant/Code/api/vendor/dingo/api/src/Provider/ApiServiceProvider.php on line 97 and defined in Handler.php line 53
```

A while ago I also bumped into this problem in my own service providers and asked Taylor about it:

```
14:20 < AntonioRibeiro> TaylorOtwell: I'm having to call mergeConfigFrom from register instead of boot, because the package need to have access to (merged) config during register. Do see any problems with this?
14:21 < TaylorOtwel> | nope
14:21 < TaylorOtwel> | i think that's fine
14:21 < TaylorOtwel> | config is unique in that regard
14:21 < TaylorOtwel> | in that it can be done from register
14:21 < TaylorOtwel> | since its loaded very early
14:21 < TaylorOtwel> | in a bootstrapper
```

### DIFF
--- a/src/Provider/ApiServiceProvider.php
+++ b/src/Provider/ApiServiceProvider.php
@@ -20,8 +20,6 @@ class ApiServiceProvider extends ServiceProvider
      */
     public function boot()
     {
-        $this->setupConfig();
-
         $this->app->singleton('Illuminate\Contracts\Debug\ExceptionHandler', function ($app) {
             return $app['api.exception'];
         });
@@ -37,6 +35,7 @@ class ApiServiceProvider extends ServiceProvider
      */
     public function register()
     {
+        $this->setupConfig();
         $this->setupClassAliases();
 
         $this->registerExceptionHandler();


GrahamCampbell:
This is wrong.

GrahamCampbell:
Packages MUST do it on boot/

GrahamCampbell:
Otherwise the entire point of that method is floored.

antonioribeiro:
The whole point of that method is not loading and merge configs? 

The problem is that `$this->registerExceptionHandler();` is using the config, which was not loaded yet, because loading will occur in boot, wayyyyyyyyyyy after register!

```
$config = $app['config']['api'];
```

GrahamCampbell:
> The whole point of that method is not loading and merge configs?

The point of doing this in the boot method it to fill out all places the user hasn't configured, NOT to provide default for the user to add to.

antonioribeiro:
And why cannot that be done on register? The user config file will be already there...

GrahamCampbell:
> And why cannot that be done on register? The user config file will be already there...

No, not necessarily.

antonioribeiro:
You mean user config may come in between register and boot calls? I would love to see this process explained...

Anyways... I had this problem once, moved to register, tested user and default config merging, and it all worked for me, and then I asked Taylor about it and he said it was ok.

If you feel it's wrong, you just have to ignore this pull request. And I just have to copy the api.php to my application config folder, no big deal.

GrahamCampbell:
https://github.com/laravel/laravel/blob/master/app/Providers/ConfigServiceProvider.php#L16

jasonlewis:
The config is being used inside the registration so it shouldn't be an issue.

Can you give me any more information as to when and how you're getting this error @antonioribeiro?

antonioribeiro:
@jasonlewis, I basically 

1) installed it and on a Laravel 5.1 app, 

2) enabled `'Dingo\Api\Provider\LaravelServiceProvider',` on my app.php

3) added `$api = app('api.router');` to the top of my routes file

4) And when I tried to hit one endpoint I got



To solve I just have to publish the config or move that method to register().

Cheers!

antonioribeiro:
This is probably [the same problem](https://github.com/dingo/api/issues/398#issuecomment-104189007) @argia-andreas found in his installation.

antonioribeiro:
Just debugged it to understand the problem: if you hit a inexistent route it will try to raise a NotFoundHttpException, so somehow it has to hit the 'api.exception' closure before booting the ServiceProvider, but the config is not fully loaded at that point yet, causing that second exception.

jasonlewis:
Can't reproduce this. Published configuration or not?

PR #454, opened 2015-06-06
fix paginator reference

tymondesigns:
rework of #448 

targeting master now

### DIFF
--- a/src/Http/Response/Factory.php
+++ b/src/Http/Response/Factory.php
@@ -7,8 +7,8 @@ use ErrorException;
 use Illuminate\Support\Str;
 use Dingo\Api\Http\Response;
 use Illuminate\Support\Collection;
-use Illuminate\Pagination\Paginator;
 use Dingo\Api\Transformer\Factory as TransformerFactory;
+use Illuminate\Contracts\Pagination\Paginator;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 
 class Factory
@@ -108,10 +108,10 @@ class Factory
     /**
      * Bind a paginator to a transformer and start building a response.
      *
-     * @param \Illuminate\Pagination\Paginator $paginator
-     * @param object                           $transformer
-     * @param array                            $parameters
-     * @param \Closure                         $after
+     * @param \Illuminate\Contracts\Pagination\Paginator $paginator
+     * @param object                                     $transformer
+     * @param array                                      $parameters
+     * @param \Closure                                   $after
      *
      * @return \Dingo\Api\Http\Response
      */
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
Surely this isn't correct. We should be typhinting an interface here?

GrahamCampbell:
We should be type hinting `Illuminate\Contracts\Pagination\Paginator`.

tymondesigns:
I know what you mean but..

It would have to be `Illuminate\Contracts\Pagination\LengthAwarePagination` since fractal requires that as described in my initial PR. 

And as @jasonlewis mentioned, that would remove the possibility of people using the standard `Illuminate\Pagination\Pagination` implementation.

Right?

jasonlewis:
I actually forgot about the interface. That would be better to type hint then the abstract. My bad.

tymondesigns:
No worries, Contract it is :+1: 

Might be worth mentioning in the wiki that they will not be able to use `User::simplePaginate()` for example, since that returns a standard Paginator instance.

src/Http/Response/Factory.php
```diff
@@ -7,8 +7,8 @@
 use Illuminate\Support\Str;
 use Dingo\Api\Http\Response;
 use Illuminate\Support\Collection;
-use Illuminate\Pagination\Paginator;
 use Dingo\Api\Transformer\Factory as TransformerFactory;
+use Illuminate\Contracts\Pagination\LengthAwarePaginator;
```
jasonlewis: Should be just `Paginator`.

I know Fractal needs a `LengthAwarePaginator` instance but that's assuming everyone will use Fractal. Keep in mind I'm not up to speed on all the pagination changes but this type hint seems to be limiting it to a single type.

tymondesigns: Yea, I agree it does seem restrictive, but if we use `Illuminate\Contracts\Pagination\Paginator` then fractal won't work at all.

To give some context:
```php
$users = User::paginate(); // returns Illuminate\Pagination\LengthAwarePaginator

// will throw an exception
return $this->response()->pagination($users, new UserTransformer);
```

Or am I missing a trick here?

tymondesigns: Thinking about it again.. maybe Fractal needs two illuminate paginator adapters. One for each paginator type, then at least we could get fractal to work in either case

jasonlewis: No... It shouldn't as the `LengthAwarePaginator` contract extends the `Paginator` contract. So even if you pass in an instance of `Illuminate\Pagination\LengthAwarePaginator` it should be fine as it is, through inheritance, an instance of `Illuminate\Contracts\Pagination\Paginator`.

tymondesigns: LOL.. I totally missed that the interface extends the other, Sorry about that.

All good then :+1: 

I will change it to the Paginator contract

tymondesigns:
all done

bweston92:
When will this be merged? Kind of need this now.

jasonlewis:
All merged. :smile:

PR #459, opened 2015-06-08
Take into consideration all files types and yaml

lucasmichot:
Also .yml files need to be taken into consideration

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,14 +1,12 @@
 root = true
 
 [*]
+charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
-
-[*.php]
-charset = utf-8
 indent_style = space
 indent_size = 4
 
-[composer.json]
+[*.yml]
 indent_style = space
-indent_size = 4
+indent_size = 2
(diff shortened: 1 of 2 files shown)


vinkla:
I'd argue this file should be removed all together. Since we just recently integrated both PHP CS Fixer and StyleCI. I no longer see the use of this file. It doesn't affect this package. Only for some users who have this integrated within there editors.

GrahamCampbell:
This file is probably fine to stay.

GrahamCampbell:
StyleCI itself has this file, even though StyleCI is enabled on StyleCI. :P

lucasmichot:
this are different things @vinkla , PHP-CS-Fixer primary goals is not to fix json, yaml and so on

lucasmichot:
Failing tests ? 5.1 stable ??

vinkla:
It exists for another purpose then the CS Fixer and StyleCI but it doesn't do anything for the library. It is custom setup for the developers building this package. That is my opinion.

lucasmichot:
Sure @vinkla 
I presume @jasonlewis will be glad to get only 4 spaces in composer.json files and not tabs in commits, or at least to avoid it

PR #471, opened 2015-06-18
updated the closure to be serialized. fixes #466

### LINKED ISSUE #466: [L5.1] config stops config:cache working
the api config currently has this:

```php
    'auth' => [
        'basic' => function ($app) {
            return new Dingo\Api\Auth\Provider\Basic($app['auth']);
        }
    ],
```

the closure there will break the `config:cache` from working, is there anyway you can rework the code to avoid the closure being needed, was thinking something like:

```php
    'auth' => [
        'basic' => 'Dingo\Api\Auth\Provider\Basic',
    ],
```

then the `config:cache` will work and and we can still use the speed boost :D

xLink:


tymondesigns:
This will need unserializing in the service provider

jasonlewis:
Is this something that *should* be done here as there are several other places in the configuration where you may provide closures.

GrahamCampbell:
Maybe this needs fixing at the laravel level?

jasonlewis:
Yeah maybe? See what Taylor thinks. If it's a widespread issue.

jasonlewis:
Hey @GrahamCampbell, what ever came of this? I'm not too keen on merging it here as it doesn't seem appropriate. Did anything ever get mentioned on the Laravel side of things?

PR #477, opened 2015-06-22
Added a pretty print json response.

jampot5000:
I'm not sure what the opinion on this will be. But I found use for it in my project, for standard browser viewable responses by setting this as the default and keeping json with no formatting accessible through the accept header.


tymondesigns:
There are many browser plugins that solve this issue. So this isn't really needed

jasonlewis:
I thought Chrome prettified it by default now? Regardless, there's a multitude of tools available to get pretty responses without this.

PR #488, opened 2015-06-27
support base path config/api.php

vanntian:
like laravel 5 config file

### DIFF
--- a/src/Provider/ApiServiceProvider.php
+++ b/src/Provider/ApiServiceProvider.php
@@ -75,7 +75,12 @@ abstract class ApiServiceProvider extends ServiceProvider
      */
     protected function setupConfig()
     {
-        $this->mergeConfigFrom(realpath(__DIR__.'/../../config/api.php'), 'api');
+        if(realpath(base_path().('/config/api.php')))
+        {
+            $this->mergeConfigFrom(realpath(base_path().'/config/api.php'), 'api');
+        }else{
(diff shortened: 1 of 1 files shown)


jasonlewis:
Doesn't `mergeConfigFrom` merge the two together so that the user defined configuration overwrites the config in the package (which this method supplies the path for)?

vanntian:
@jasonlewis yes

bweston92:
@jasonlewis yes that is correct, this merge request stops the app specific and default merging.

jasonlewis:
So what's the benefit here exactly...

bweston92:
I actually think that it would merge an application specific configuration with it-self, as the first argument is supposed to be the path to the package specific configuration.

jasonlewis:
Yeah I don't think this does what's intended. The idea of `mergeConfigFrom` is to merge the default configuration in the package with the application specific configuration.

natsu90:
So anyway we can have this? I just realised my Lumen doesn't pick up config from base path. I've to add `$app->configure('api')` in `bootstrap/app.php` myself.

jasonlewis:
Okay I think I see what's happening here for Lumen installs. Let me work on something.

jasonlewis:
I've made a change that automatically configures it with `$app->configure` so you don't need to worry about it. It merges the default package configuration as well.

natsu90:
Thanks.

PR #505, opened 2015-07-04
fix: adding controller multiple times

low-res:
as described in https://github.com/dingo/blueprint/pull/1 i had trouble with duplicate controllers and actions in my generated api docs. as far as i understand it, there was no check for duplicate controllers and/or actions.


src/Console/Command/Docs.php
```diff
@@ -96,7 +96,9 @@ protected function getControllers()
         foreach ($this->router->getRoutes() as $collections) {
             foreach ($collections as $route) {
                 if ($controller = $route->getController()) {
-                    $controllers[] = $controller;
+                    if(!$controllers->contains($controller)) {
```
GrahamCampbell: cs

PR #532, opened 2015-07-22
Adding the method Api\Routing\Route::isProtected() to fix the console command api:routes

yosvelquintero:


jasonlewis:
Those providers aren't a requirement for a route to be classified as protected. I'm also not going to be adding the isProtected method back.

wesrice:
@jasonlewis, I'm assuming @yosvelquintero made this pull request to remediate the following error when running `php artisan api:routes`:

```
PHP Fatal error:  Call to undefined method Dingo\Api\Routing\Route::isProtected() in [...]vendor/dingo/api/src/Console/Command/Routes.php on line 73
```

If you don't plan on adding the `isProtected` method back, how will this be remediated? Will you be removing the `protected` column from the output routes?

jasonlewis:
I understand what the issue is I'm just not in a position right now to fix
it. The fix will most likely take place in the command itself. It needs to
check if the route has the API.auth middleware attached.

PR #538, opened 2015-07-24
A route has no "isProtected" method anymore

rokde:
This change was committed with 29084209b7e5f0fdd6af50f684c5159adebfac69


jasonlewis:
I'll merge this for now so as the command works again. However, once I'm home I'll ammend this so as the protected column is still a part of the table.

Cheers.

PR #541, opened 2015-07-27
Fix ValidationHttpException.php

jszobody:
I believe this is a typo, and should be `$previous`? Otherwise I get a 500 error "Undefined variable: exception."


jszobody:
Bah, didn't notice there was already a PR for this issue. Closing.

PR #544, opened 2015-07-29
Update to accept parser so that versions can be accepted without a v …

### LINKED ISSUE #543: If you don't have the letter "v" as a prefix to your version - it breaks Dingo
Dingo will not match based on the presumed pattern for the version.

For example, if I set my api versions as 1.0, 2.0.etc. (instaed of v1.0, v2.0) then the pattern matching within the accept class fails.

kirkbushell:
Resolves #543.

### DIFF
--- a/src/Http/Parser/Accept.php
+++ b/src/Http/Parser/Accept.php
@@ -59,7 +59,7 @@ class Accept implements Parser
     {
         $default = 'application/vnd.'.$this->vendor.'.'.$this->version.'+'.$this->format;
 
-        $pattern = '/application\/vnd\.('.$this->vendor.')\.(v[\d\.]+)\+([\w]+)/';
+        $pattern = '/application\/vnd\.('.$this->vendor.')\.(v?[\d\.]+)\+([\w]+)/';
 
         if (! preg_match($pattern, $request->header('accept'), $matches)) {
             if ($strict) {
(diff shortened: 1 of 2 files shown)


bweston92:
I believe that having the `v` prefix actually lets people know it is a version number.

kirkbushell:
That may be so, but it shouldn't be a requirement. The way API versions are done with Dingo I actually find quite odd (rather than being a separate header).

It caught me up, and caused exceptions/issues that were hard to track down and debug. It's also not documented.

lightvision:
I also beleave this even somerimes I would ti have the liberty to name versions withot v.
Accepting versions without v wouldn't involve a change to api docs too?

kirkbushell:
I don't see how - it would simply mean that the package doesn't make any assumptions about how you name your versions. What if you want to use codenames instead? Like "Icecream Sandwich" :)

lightvision:
@jasonlewis What do you think about this?

@kirkbushell I got your point. I was annoyed too about forcing me to put a v before the actual version, but I had to move over for the moment, this is something I can leave with.

jasonlewis:
I'm fine with this. I guess it's a silly limitation. As for the header, I personally don't like making a new request header. Content negotiation is best done with the `Accept` header. I really need to document vendor prefixes a bit better though as *technically* they need to be registered with the IANA. I want to tweak content negotiation a bit.

PR #557, opened 2015-08-04
CS Fixes

GrahamCampbell:
This makes sure your actual coding style is enforced. You seem to be following many more rules than just PSR-2. Let me know if I've enabled a fixer you didn't want by mistake.

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,4 +1,12 @@
-preset: psr2
+preset: recommended
+
+linting: true
 
 enabled:
-  - short_array_syntax
+  - logical_not_operators_with_successor_space
+
+disabled:
+  - align_double_arrow
+  - new_with_braces
+  - ordered_use
+  - psr0
--- a/config/api.php
+++ b/config/api.php
@@ -123,7 +123,7 @@ return [
         'errors' => ':errors',
         'code' => ':code',
         'status_code' => ':status_code',
-        'debug' => ':debug'
+        'debug' => ':debug',
     ],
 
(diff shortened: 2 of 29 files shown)


jasonlewis:
Where does the comma after the final array come into play? Looks gross IMO. :grin:

jasonlewis:
I think the only thing I do is `if (! $whatever)`, where there's a space after the exclamation mark. I'm not sure PSR-2 makes any recommendations there do they?

Where do I not follow PSR-2?

GrahamCampbell:
> Where does the comma after the final array member come into play? Looks gross IMO.

Sure I can disable it. You just have some places where you have trailing commas, and there's no fixer to remove them. The main advantage to having trailing commas is regarding only changing one line when adding to an array thus reducing merge conflicts and having clearer diffs.

> I think the only thing I do is if (! $whatever), where there's a space after the exclamation mark. I'm not sure PSR-2 makes any recommendations there do they?

Yeh, that's allowed in PSR-2. I recently enabled that fixer on laravel/framework.

> Where do I not follow PSR-2?

You do follow PSR-2. What I meant is you also have additional rules that you're using on top of PSR-2, but you weren't actually enforcing them before.

jasonlewis:
I'll just have to get over the ugliness of it. Won't take me long. :stuck_out_tongue:

jasonlewis:
Cheers @GrahamCampbell! :+1:

GrahamCampbell:
No problem. :)

PR #581, opened 2015-08-16
added support for group 'as' similar to illuminate router

pulkitjalan:
https://github.com/illuminate/routing/blob/master/Router.php#L389

### DIFF
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -434,7 +434,11 @@ class Router
 
         $new['where'] = array_merge(array_get($old, 'where', []), array_get($new, 'where', []));
 
-        return array_merge_recursive(array_except($old, ['namespace', 'prefix', 'where']), $new);
+        if (isset($old['as'])) {
+            $new['as'] = $old['as'].array_get($new, 'as', '');
+        }
+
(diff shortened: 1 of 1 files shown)


jasonlewis:
So this behaviour seems a bit odd (even the Laravel one). It just adds them together with no separator.

So:

```php
$api->version('v1', ['as' => 'api'], function ($api) {
    $api->get('users', ['as' => 'users', function () {
        //
    }); 
});
```

The name for the `/users` route is `apiusers`. I would've imagined it would be `api.users`. Same goes for Laravel though. There's no `.` separating them. I've never used the `as` in a group before but this is expected behaviour?

pulkitjalan:
Yea this is expected, same with Laravel, you need to set the separator.

```php
$api->version('v1', ['as' => 'api.'], function ($api) {
    $api->group(['prefix' => 'users', 'as' => 'users.'], function ($api) {
        $api->get('', ['as' => 'index', function () {
            //
        });
        $api->post('', ['as' => 'store', function () {
            //
        });
    });
});
```
Not the name for `GET /users` route is `api.users.index` and `POST /users` is `api.users.store`

jasonlewis:
Ah righto then, no worries. As long as that's the expected behaviour! :smile:

Issue #590, opened 2015-08-19
API Blueprint: repeated doc generation for Resource


yuters:
When you have `$api->resource('resource', 'ResourceController');` in your router file, the documentation for this particular resource is generated multiple times.

PR #591, opened 2015-08-19
Fix issue #590 repetition in blueprint generation.

yuters:
The contains() method was not entirely working with the $controller
object. By adding the class name as a key in the controllers Collection,
The contains() method looks at that string instead.


jasonlewis:
I actually looked at the source after a merge. `contains` does an `in_array` so it should have been looking at the value. I'll update this further to use `has` instead of `contains` so we now check the key instead of the value.

Not sure why you were getting duplicates in your controllers. But hopefully this will eventually resolve it.

PR #622, opened 2015-09-01
Added `any` and `match` router functions

pulkitjalan:
Matching laravel's router.

https://github.com/laravel/framework/blob/5.1/src/Illuminate/Routing/Router.php#L195


pulkitjalan:
added tests

Issue #640, opened 2015-09-15
api:cache not working when only using dingo router


eduacostam:
`api:cache` command calls Laravel's `route:cache` command, which in turn generates the `app/bootstrap/cache/routes.php` file with the following header:
```php
<?php

/*
|--------------------------------------------------------------------------
| Load The Cached Routes
|--------------------------------------------------------------------------
|
| Here we will decode and unserialize the RouteCollection instance that
| holds all of the route information for an application. This allows
| us to instantaneously load the entire route map into the router.
|
*/
```

In my case I'm not using Laravel router as all routes in my application are API specific. This in turn makes the call to `route:cache` do nothing as there are no known routes for laravel router.
The thing is that Dingo's `api:cache` command appends to the generated file, this makes the generated file be an invalid PHP file as it does not have the PHP open tag.

PR #641, opened 2015-09-15
Fix issue #640 api:cache command not working when only using dingo router

eduacostam:
Issue is explained in #640.
If no route file was generated by `route:cache` command then prepend PHP open tag to route cache file stub.

### DIFF
--- a/src/Console/Command/Cache.php
+++ b/src/Console/Command/Cache.php
@@ -65,9 +65,14 @@ class Cache extends Command
         }
 
         $stub = "app('api.router')->setAdapterRoutes(unserialize(base64_decode('{{routes}}')));";
+        $path = $this->laravel->getCachedRoutesPath();
+
+        if (! $this->files->exists($path)) {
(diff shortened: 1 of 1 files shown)


src/Console/Command/Cache.php
```diff
@@ -65,9 +65,14 @@ public function handle()
         }
 
         $stub = "app('api.router')->setAdapterRoutes(unserialize(base64_decode('{{routes}}')));";
+        $path = $this->laravel->getCachedRoutesPath();
+
+        if (! $this->files->exists($path)) {
+            $stub = "<?php\n\n".$stub;
```
GrahamCampbell: why not just `$stub = "<?php\n\n$stub";`

eduacostam: Yeah... that might be better, don't know why it didn't occur to me. Going to fix it now.

eduacostam: Done! Thanks!

jasonlewis:
Many thanks.

PR #658, opened 2015-09-24
Fix doblocks

lucasmichot:
Fix doblocks


lucasmichot:
https://styleci.io/analyses/Xly9dq is a false positive.
See https://github.com/lucasmichot/api/blob/feature/master.doblocks/src/Routing/Router.php#L563

lucasmichot:
ping @GrahamCampbell : check https://styleci.io/analyses/Xly9dq

PR #661, opened 2015-09-24
Call directly functions form Arr and Str

lucasmichot:
Call directly functions form Arr and Str

### DIFF
--- a/src/Console/Command/Docs.php
+++ b/src/Console/Command/Docs.php
@@ -4,6 +4,7 @@ namespace Dingo\Api\Console\Command;
 
 use ReflectionClass;
 use Dingo\Blueprint\Writer;
+use Illuminate\Support\Arr;
 use Dingo\Api\Routing\Router;
 use Dingo\Blueprint\Blueprint;
 use Illuminate\Console\Command;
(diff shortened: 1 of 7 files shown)


jasonlewis:
Unless you can provide a really good reason as to why I shouldn't use the functions I don't see a point in merging this.

lucasmichot:
Hey @jasonlewis 
The purpose is to avoid unnecessary calls, in the `array_*` and `string_*` helper functions case, we call and helper function which is doing nothing but calling the real `\Suppor\Arr::` or `\Suppor\Str::` function.
If the helper function was doing something on the input or the output, it would make sense.
But in this situation not; it is making things slower (and some might think a bit less OO too)

jasonlewis:
What sort of overhead are we talking about exactly?

lucasmichot:
Calls to `array_` and `str_` helpers functions have been removed in favor of `\Suppor\Arr::` or `\Suppor\Str::` functions in `laravel/framework`. The functions are directly called without using the helper function (a function calling another function).
Was it the meaning of your question ?

pulkitjalan:
Calling the function directly makes more sense anyway right, instead of the helper?

Even if the overhead is not that high it still would be less efficient to call the helper which would just call the function anyway?

tymondesigns:
The only plausable reason to change this imo, would be to avoid potential clashes with other function declarations of the same name. But since this is a laravel specific package, that doesn't apply. 

I also think any performance consequences of calling a function that calls a method are negligible, especially given the many micro optimisations that could also be made elsewhere in a php application.

Aside from that I do prefer to use the classes directly though, as it feels a little cleaner. Just my thoughts :)

lucasmichot:
Both you guys @pulkitjalan and @tymondesigns are right indeed.

jasonlewis:
I cave.

:wink:

PR #668, opened 2015-09-29
Repair undefined

a939638621:
Sometimes unable to read configuration , display undefined


jasonlewis:
Same as other PR. Set this yourself, either through the config or environment variables.

a939638621:
@jasonlewis  I have already written in the env file to this configuration, but sometimes there will be no definition, so I tried to change the source code, the results are not in this problem！！！

I can't speak English, these are the machines!! I hope!!

PR #669, opened 2015-09-29
Update api.php

a939638621:


jasonlewis:
Nope.

This is up to you to set. I won't provide a default.

PR #693, opened 2015-10-19
Allow includes for non-collections

headcastlab:
On trying to created an include for datestamps via a date transformer, I found that eager loading always looks for a model relation. Now returns false if the include is not a collection.

### DIFF
--- a/src/Transformer/Adapter/Fractal.php
+++ b/src/Transformer/Adapter/Fractal.php
@@ -113,6 +113,8 @@ class Fractal implements Adapter
     {
         if ($response instanceof IlluminatePaginator) {
             $response = $response->getCollection();
+        } else {
+            return false;
         }
 
         return $response instanceof EloquentCollection && $this->eagerLoading;


jasonlewis:
Hm, I don't like simply returning `false` like that. Doesn't seem right.

What exactly is this datestamp and what is the response you're trying to transform?

headcastlab:
The datestamp is a carbon instance from the eloquent created_at / updated_at db columns which I want to include optionally and transform. As it is not a collection and also would not require any eager loading my solution was simply to return false.

jasonlewis:
But why was it being picked up as an instance of a `IlluminatePaginator` in the first place? And if it's not then it's being picked up as an `EloquentCollection` then.

headcastlab:
That sir is something I am unable to comment on... :)

My code was: I have a DateTransformer which I use to selectively include dates as includes from other transformers like so:

    public function includeCreated(Model $model)
    {
        return $this->item($model->created_at, new DateTransformer);
    }

Which triggers the error:

Missing argument 1 for Illuminate\Database\Eloquent\Model::created(), called in /home/*****/api/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php on line 477 and defined

Which suggested that the include was trying to load a models relation. From there I hacked and slashed code until I figured out out to make it work :shame:

headcastlab:
I have hit another use case where eager loading is causing the same issue. I am passing a custom collection to a transformer which is not a relation...

The Product Model has these relations:

```    /**
     * Get all the characters that are assigned this product
     * @return [type] [description]
     */
    public function characters()
    {
        return $this->morphedByMany('App\Models\Character', 'productable');
    }    
   
    /**
     * Get all the stories that are assigned this product
     * @return [type] [description]
     */
    public function stories()
    {
        return $this->morphedByMany('App\Models\Story', 'productable');
    }

**Comment by headcastlab** [comment]
Sorry, click happy closed...

Another though... can I selectively disable eager loading?

**Comment by jasonlewis** [comment]
You can disable/enable it at runtime now (pull latest changes).

With an instance of the transformer factory grab the adapter and call `disableEagerLoading()`.

```php
app('api.transformer')->getAdapter()->disableEagerLoading();
```

I also added `__call` functionality to the factory, so you can call it directly also.

```php
app('api.transformer')->disableEagerLoading();
```

headcastlab:
Ace, thank you for your help and your excellent work.

PR #695, opened 2015-10-21
fixed scopes except not working as expected

pulkitjalan:
```
$this->scopes('foo', ['except' => 'index']);
```

The above statement should still add the scopes to all other methods, however currently it does not. this should fix.

Also added test.

### DIFF
--- a/src/Routing/Route.php
+++ b/src/Routing/Route.php
@@ -203,8 +203,8 @@ class Route
             return true;
         } elseif (isset($options['only']) && in_array($this->method, $this->explodeOnPipes($options['only']))) {
             return true;
-        } elseif (isset($options['except']) && in_array($this->method, $this->explodeOnPipes($options['except']))) {
-            return false;
+        } elseif (isset($options['except'])) {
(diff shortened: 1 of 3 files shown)


src/Routing/Route.php
```diff
@@ -203,8 +203,12 @@ protected function optionsApplyToControllerMethod(array $options)
             return true;
         } elseif (isset($options['only']) && in_array($this->method, $this->explodeOnPipes($options['only']))) {
             return true;
-        } elseif (isset($options['except']) && in_array($this->method, $this->explodeOnPipes($options['except']))) {
-            return false;
+        } elseif (isset($options['except'])) {
+            if (in_array($this->method, $this->explodeOnPipes($options['except']))) {
```
lucasmichot: Why not just `return ! in_array($this->method, $this->explodeOnPipes($options['except']));` ?

pulkitjalan: Good point, changing.

pulkitjalan: Changed, Thanks.

jasonlewis:
Nice one. Cheers.

Issue #692, opened 2015-10-19
validate a route scopes bug


argb:
This method "protected function validateRouteScopes(AccessTokenEntity $token, Route $route)" in about line 118 of Provider/OAuth2.php ,is wrong logic.
it should be all scopes satisfy the requirement ,not any of scope exist.
maybe the right codes are : 
$scopes = $route->scopes();

protected function validateRouteScopes(AccessTokenEntity $token, Route $route)
    {
        $scopes = $route->scopes();

        if (empty($scopes)) {
            return true;
        }

        foreach ($scopes as $scope) {
            if (!$token->hasScope($scope)) {
                throw new InvalidScopeException($scope);
            }
        }
        return true;
    }

pulkitjalan:
Agreed, created pull request #698

argb:
@pulkitjalan Just from this bug i can assert Dingo api should have many bugs,so obvious bug should not be occur.

jasonlewis:
This was actually decided on a long time ago because it's very implementation specific. (See #146)

It initially required ALL scopes to be present for it to validate. I'm happy to have another discussion about it. But don't assume it's a bug just because it's not behaving how you think it should behave. Again, you might want a different implementation than someone else.

Best course of action here would be to make it configurable I suppose.

> Just via this bug i can assert Dingo api should have many bugs,so obvious bug should not occur.

If that's what you believe @argb then don't use the package. It's pretty simple.

pulkitjalan:
Fair enough, do you think configurable via the config file or configurable per route?

pulkitjalan:
I have made what I think should work and added tests. Can you take a look before I submit a pr. https://github.com/pulkitjalan/api/commit/481c9ad3ff3a322a0e3d8cda68257e63b85ac51c. It adds a new option to the routes `requireAll` which is used to validate if all scopes are required or defaults to false which is any scopes are required.

jasonlewis:
I like the implementation but I think `requireAll` is a little vague.

Other possibilities:

- `scopeAll`
- `strict` (also vague, already have a `strict` config option for Accept header)
- `scopeStrict`

I hate naming things. Definitely need some input for the name.

pulkitjalan:
How about scopeStrict to some what match the strict accept header functionality?

jasonlewis:
Okay, submit the PR and we'll move this discussion there. :+1:

argb:
[...]

PR #698, opened 2015-10-23
changed oauth2 logic to require all scopes instead of any

pulkitjalan:
Changed OAuth2 logic to require all scopes instead of any as mentioned in #692


jasonlewis:
This isn't a bug. We'll discuss this further in the issue.

PR #700, opened 2015-10-26
Add properties doc for magic-method properties in Helpers trait.

ryanwinchester:
Helps IDEs, mostly, and those of us who use them.

I'm assuming there's a reason you use `\Illuminate\Auth\GenericUser|\Illuminate\Database\Eloquent\Model` instead of just `\Illuminate\Contracts\Auth\Authenticatable` so I left it at that.

### DIFF
--- a/src/Routing/Helpers.php
+++ b/src/Routing/Helpers.php
@@ -4,6 +4,12 @@ namespace Dingo\Api\Routing;
 
 use ErrorException;
 
+/**
+ * @property \Dingo\Api\Dispatcher                                            $api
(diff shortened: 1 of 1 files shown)


ryanwinchester:
phpstorm

ryanwinchester:
phpstorm 1

joshhornby:
@ryanwinchester Slightly unrelated but what theme is that?

ryanwinchester:
@joshhornby https://github.com/ChrisRM/material-theme-jetbrains

jasonlewis:
I'd say I'm not using `Authenticatable` because that method hasn't changed for a *long* time. There's still a few odds and ends that could (should) be updated, mainly in docblocks.

PR #707, opened 2015-10-30
Allow later versions of Fractal

aejnsn:
Fractal 0.12.0 is the only version accepted. We use 0.13.0. If Dingo doesn't have any need for 0.12 specifically--if so, I'd like to know which parts need to be updated--then 0.12.0 or higher should be here. Hopefully Fractal is following SemVer and the minor update won't break anything.


aejnsn:
I just ran PHPUnit local, and it didn't fail. Why is it failing on Travis?

PR #711, opened 2015-11-04
Configurable oauth scope

pulkitjalan:
Made requiring all scopes or any scopes configurable as discussed in #692


jasonlewis:
I'm happy with this. Keeps the current default and allows you to configure it with `scopeStrict` on the route itself. Thanks @pulkitjalan for your work on this.

Issue #699, opened 2015-10-23
Disable logging of exceptions in production environment


fabianlaule:
Currently, every exception will be logged in the logfile which is unnecessary. Unfortunately, there is no possibility to disable it. 
[This method](https://github.com/dingo/api/blob/51901c3dda0ccf099a8a2515141e5a0fdbd402bf/src/Exception/Handler.php#L123) should not be called on HttpKernel or Dingo specific exceptions to prevent a fast growing log file with unnecessary entries.

An example of a thrown and logged exception:
```
production.ERROR: exception 'Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException' with message ...
Stack trace: 
....
```

jasonlewis:
No I think it's more than that. The default kernel has the ability to not report certain exceptions, which I think `HttpException` is a part of. So, by default, none of those exceptions are being reported. I'll need to mimic this functionality as best I can, as errors still need to be reported, just not all of them.

alextyrey:
Was going to submit a PR that addresses this but..

@jasonlewis is there a reason Dingo's exception handler can't just copy the functionality of Laravel's $dontReport?

I feel like your above comment implies that the issue is more complex than I think it is.

lvandyk:
Any update on this? Also needed to disable logging entirely on production for now.

jasonlewis:
A fix for this will be pushed in the next update.

lvandyk:
Awesome thanks Jason!

fabianlaule:
Perfect, thanks Jason :+1:

PR #726, opened 2015-11-13
add dontReport functionality

jarbitlira:
I tried to add the Exceptions to the $dontReport array  like say laravel documentation and this not work.

So, I did these changes to the core of dingo/api, I tested this something fast but I think that it's fine.

I believe that this is related with the issue #699

### DIFF
--- a/config/api.php
+++ b/config/api.php
@@ -206,4 +206,6 @@ return [
 
     ],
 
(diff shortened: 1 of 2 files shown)


jarbitlira:
Enhancing and doing testing

Issue #747, opened 2015-11-30
[Proposal] Automatically exclude `create` and `edit` from resource routes?


HipsterJazzbo:
I find that when I'm building `$api-resource()` routes, i nearly always end up passing the option `['except' => ['create', 'edit']]`, because I don't think they make any sense for an API.

Is it worth excluding those by default for API resources? Would certainly make my routes file a lot cleaner.

jasonlewis:
I'd be happy to accept a PR for this, because yes, you're right, they're essentially pointless routes in terms of an API.

PR #749, opened 2015-12-01
Exclude 'create' and 'edit' from resource routes

HipsterJazzbo:
See #747.


jasonlewis:
I sure hope this doesn't upset someones code. I'd be worried if it did! Thanks @HipsterJazzbo!

Issue #751, opened 2015-12-03
Rate Limiting should not apply to Internal Requests


aleemb:
Not sure if it makes sense to count internal API requests toward the `limit`.

lightvision:
From my point of view it make sense. If you are using your own api through internal requests in order to deploy saas apps it sure make sens. Think about you have several plans based on api usage ...

aleemb:
I believe the more common use-case is to not limit web-form usage (Twitter, Google Search, Analytics, etc). In fact, I cannot think of any service that limits web-form usage along with API usage. So the limit is imposed on direct API calls only. In the interest of sensible defaults, it would be better to cater to the more common use-case.

Hopefully, a few others can chime in with their thoughts? @jasonlewis @lucasmichot @vinkla

lucasmichot:
@aleemb indeed, it does not make sense with an internal request.
A good default would be to skip the rate limiting in this case.

Just my opinion -- @jasonlewis @vinkla ?

vinkla:
I agree with @lucasmichot. Since you have full control over the internal requests anyway.

andreygrehov:
I partially agree. Internal API consuming should be configurable, i.e. one can either rate limit it or disable the rate limiting all together. The reason for that is because internal API requests may potentially be more harmful than external ones.

jasonlewis:
#752 was merged.

andreygrehov:
thank you, @jasonlewis. It's an off-topic, but would be great to plan the 1.0 release, unless it hasn't been yet.

PR #752, opened 2015-12-03
Do not throttle internal requests

lucasmichot:
This PR targets #751 .
@aleemb @jasonlewis @vinkla thoughts and suggestions are welcome

### DIFF
--- a/src/Http/Middleware/RateLimit.php
+++ b/src/Http/Middleware/RateLimit.php
@@ -5,6 +5,7 @@ namespace Dingo\Api\Http\Middleware;
 use Closure;
 use Dingo\Api\Http\Response;
 use Dingo\Api\Routing\Router;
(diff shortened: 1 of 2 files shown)


vinkla:
LGTM.

aleemb:
Looking good, that was quick. Thanks much!

Aleem

jasonlewis:
I wouldn't hand off to the next request until we confirm the request is indeed internal. Just run the `$next($request)` twice, it's no biggie.

Otherwise the rate limiting is pretty much void as you're essentially firing off the request by handing it off to the next middleware (until the route itself is fired).

lucasmichot:
@jasonlewis , I am not sure what you mean ?
"hand off" ?

lucasmichot:
OK, I see. NVM
See: 211d4179a6e3908795380b2803051f5dd011e314

PR #760, opened 2015-12-08
Update domain validation ignoring ports.

### LINKED ISSUE #754: Issues With Ports
There are a couple different issues with ports that have been brought up before, specifically the last comment in #570.

I ran across this same error and believe I have found the culprit, but want to make sure I'm not doing anything funky.

The domain of the request is validated twice, once by Laravel and once by Dingo. Unfortunately, the validators work differently. Specifically, the Dingo validator(Dingo\Api\Http\Validation\Domain) compares the host header, including port number. The laravel validator(Illuminate\Routing\Matching\HostValidator) on the other hand uses `Request::getHost` which excludes the port.

You should be able to fix the issue by switch your validation to wrap the laravel validator or use `getHost()`.

MTon:
fixes #754

### DIFF
--- a/src/Http/Validation/Domain.php
+++ b/src/Http/Validation/Domain.php
@@ -7,6 +7,8 @@ use Dingo\Api\Contract\Http\Validator;
 
 class Domain implements Validator
 {
+    const PATTERN_STRIP_PROTOCOL = '/:\d*$/';
+
     /**
      * API domain.
      *
@@ -35,7 +37,7 @@ class Domain implements Validator
      */
     public function validate(Request $request)
     {
-        return ! is_null($this->domain) && $request->header('host') == $this->stripProtocol($this->domain);
+        return ! is_null($this->domain) && $request->getHost() == $this->getStrippedDomain();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Http/Validation/Domain.php
```diff
@@ -53,4 +55,32 @@ protected function stripProtocol($domain)
...
+    /**
+     * Strip the port from a domain.
+     *
+     * @param $domain
+     *
+     * @return mixed
+     */
+    protected function stripPort($domain)
+    {
+        $domainStripped = preg_replace(self::PATTERN_STRIP_PROTOCOL, '', $domain);
```
lucasmichot: I would just do:
```php
if ($domainStripped = preg_replace(self::PATTERN_STRIP_PROTOCOL, null, $domain)) {
    return $domainStripped;
}

return $domain;
```

MTon: Yes, you're right, more elegant! Thx

lucasmichot: But in fact, I should have been a bit more verbose in my previous comment... does it really makes sense to have `$domainStripped`.
We coud achieve exactly the same doing:
```php
return preg_replace('/:\d*$/', null, $domain);
```

MTon: No in case of an error 

```php
if ($domainStripped = preg_replace(self::PATTERN_STRIP_PROTOCOL, null, $domain)) {
    return $domainStripped;
}

return $domain;
```

the return value would be the given domain, not null.

lucasmichot: This could only occur on malformed domain, strange we don't already have a failing test for that

lucasmichot: Nevertheless this is the role of the regexp itself to handle this case.
IMO, `preg_match` is more appropriate and a failing test should be added.

lucasmichot: Something like this:
```php
preg_match('/(.*):\d+/', $domain, $matches);

return isset($matches[1]) ? $matches[1] : $domain;
```

jasonlewis:
Thanks mate! :smile:

PR #764, opened 2015-12-11
Separate out Json and JsonAPI formatters

JamesGuthrie:
For some applications it may be favourable to provide a simple JSON api,
one which returns objects and arrays instead of the structured jsonapi
approach. The new SimpleJson class does not wrap eloquent models or
collections with keys.

I haven't added documentation for it yet, but I can do so if you would would
like to merge this PR.


jasonlewis:
Not looking for anything like this to be merged in. Folks can pull in other packages that provide different formatters should they need them.

PR #773, opened 2015-12-16
Scope Identifiers not getting passed down to the Fractal Scope

theHarvester:
- Extracted new key "scopeIdentifier" from the parameters passed into the binding
 - Pushed scopeIdentifier into the fractal scope instantiation


jasonlewis:
Have changed it to just `identifier` as well to simplify things.

PR #774, opened 2015-12-17
Slack API room support

lucasmichot:
@jasonlewis , because many issues are in fact support questions.


jasonlewis:
Are we allowed to just jump in on that room? I mean, I don't care, I don't mind answering questions here either. I tend to forget about slack unless some @ mentions me. :grin:

PR #776, opened 2015-12-18
Move the SetupConfig method to register and not in boot

joselfonseca:
when running a test in an application i notice that the config is being
register in the boot method and not in the register of the service
provider, this causes that when the register exception handler method
is called, the config is null since the boot method has not been called
yet. I discover this in a test that i was writing for my application
and i got this
```
Hechoenlaravel\JarvisFoundation\Tests\EntityEntries\TestEntriesCommands::test_it_validates_an_entry_for_entity_and_throws_exception_if_missing_data
ErrorException: Argument 2 passed to
Dingo\Api\Exception\Handler::__construct() must be of the type array,
null given, called in
/home/vagrant/Code/jarvis-platform/packages/hechoenlaravel/jarvis-founda
tion/vendor/dingo/api/src/Provider/ApiServiceProvider.php on line 125
and defined 
```
after some debugging i was able to find that.

### DIFF
--- a/src/Provider/ApiServiceProvider.php
+++ b/src/Provider/ApiServiceProvider.php
@@ -23,8 +23,6 @@ abstract class ApiServiceProvider extends ServiceProvider
      */
     public function boot()
     {
-
-
         Http\Response::setFormatters($this->prepareConfigValues($this->app['config']['api.formats']));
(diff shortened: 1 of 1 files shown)


jasonlewis:
This has been done, cheers.

PR #797, opened 2016-01-05
Ability to add totally custom exception handler

mikkokut:
Config option to be able to implement totally custom exception handler which extends from the <code>\Dingo\Api\Exception\Handler</code>. 

Added config option <code> exceptionHandler</code> which takes class for the handler. Defaults to original implementation.


0plus1:
This is a very welcome change!

driesvints:
:+1:

marcorivm:
:+1:
Needing this feature to allow Errors to be logged to external services like Bugsnag

lucasmichot:
How do you ensure your custom class constructor expects these arguments ?
At minima any custom exception handler class should implément an interface defining the __construct method

driesvints:
@lucasmichot unfortunately the Dingo handler requires 2 configuration settings in order to properly function. Since Laravel's IoC container has no way of auto-resolving those, we're a bit stuck with the way Dingo instantiates this class. To be honest: as long as you're not using external dependencies in the handler I think it's ok to just extend the Dingo handler and overwrite the `report` and `handle` methods to do what you need your handler to do.

If you *do* however need external dependencies (say Bugsnag or w/e), you could overwrite Dingo's service provider with this approach:

```php
    protected function registerExceptionHandler()
    {
        $provider = $this->app['config']['api']['exceptionHandler'] ?? ExceptionHandler::class;

        $this->app->singleton('api.exception', $provider);
    }
```

Or the PHP 5.x version :smile: 

```php
    protected function registerExceptionHandler()
    {
        if (isset($this->app['config']['api']['exceptionHandler'])) {
            $provider = $this->app['config']['api']['exceptionHandler'];
        } else {
            $provider = ExceptionHandler::class;
        }

        $this->app->singleton('api.exception', $provider);
    }
```

And than replace the Dingo exception handler construct with something like this:

```php
    /**
     * Create a new exception handler instance.
     *
     * @param \Psr\Log\LoggerInterface $log
     * @param array                    $format
     * @param bool                     $debug
     *
     * @return void
     */
    public function __construct(LoggerInterface $log, ApiConfig $config)
    {
        $this->log = $log;
        $this->format = $config->format();
        $this->debug = $config->debug();
    }
```

With `ApiConfig` being a custom class for Dingo which uses Laravel's config store to get the config variables. Watch out with this approach though because this class may be abused :smile: 

Or you know,... just use `app()` ¯\\\_(ツ)_/¯

> Side note: Excellent read on configuration variables as class dependencies: http://paul-m-jones.com/archives/6203

hskrasek:
Something I am noticing with this PR is that even if the custom class is specified, the report function is not being called automatically. Currently getting around this by overriding the binding of api.exception and calling report from the handle method.

abcsun:
:+1: it is great to the custom exceptionHandler

jasonlewis:
Some changes were made to exception handling in ba036c01a72e79d4a3b0a19c1ec8f0590830bc19.

You should also be able to simply rebind if you need to configure it further. Does it really need its own configuration option?

driesvints:
@jasonlewis no, this is even better. Thanks! :)

mikkokut:
@jasonlewis  Yes, you are right. Rebinding is much better approach.

PHPisGod:
@jasonlewis 
$api->group(['middleware'=>'api.throttle','limit'=>5,'expires'=>1],function($api){}
I want to limit the api frequency ,so I use this middleware and it works 
but I want to change the default respon 
I know that I can change the respon key in api.php but I can't change the respon content 
now:
{
  "message": "You have exceeded your rate limit.",
  "code": 403
}
what I want
{
  "message":"You can't make fun of me!"
  "code":213
}

pcastrotigre:
Hi @jasonlewis How should I rebind the handler? In my AppServiceProvider I am doing this

```
$this->app->bind('api.exception', function ($app) {
    $config = $app['config']['api'];
    return new \App\Exceptions\ApiExceptionHandler($app['Illuminate\Contracts\Debug\ExceptionHandler'], $config['errorFormat'], $config['debug']);
});
```
And my class is:
```
namespace App\Exceptions;

use Dingo\Api\Exception\Handler;

class ApiExceptionHandler extends Handler
{
}
```
However I am receiving the error 

**Fatal error: Maximum function nesting level of '256' reached, aborting! in /home/vagrant/smart-office-api/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php on line 115**

hskrasek:
@pcastrotigre I'm using the following method to rebind the handler:

```php
        $this->app->singleton('api.exception', function ($app) {
            $config = $app['config']['api'];

            return new Handler(
                $app['Illuminate\Contracts\Debug\ExceptionHandler'],
                $config['errorFormat'],
                $config['debug']
            );
        });
```

roberttolton:
Is there a full example implementation of this anywhere?

TheGeekyM:
In your AppServiceProvider add this:

`$this->app->bind('api.exception', function ($app) {
           return new Handler($app);
        });`

PR #800, opened 2016-01-06
Fixed 'where' attribute for group() method.

Roquie:
Bug example:
```php
$wheres = [
     'foo_id' => '[0-9]+'
];
$router->group(['version' => 'v1', 'prefix' => 'v1', 'where' => $wheres], function($router) {
        $router->get('/url/{foo_id}', function() {
             return "you don't need to see this";
        });
});
```
So, we open browser and go to link:
`http://example.com/v1/url/<asdasd>`
and see it:
```
you don't need to see this
```
but should:
```json
{
    "message": "404 Not Found",
    "status_code": 404
}
```

### DIFF
--- a/src/Routing/Adapter/Laravel.php
+++ b/src/Routing/Adapter/Laravel.php
@@ -113,6 +113,7 @@ class Laravel implements Adapter
         $this->createRouteCollections($versions);
 
         $route = new Route($methods, $uri, $action);
+        $route->where($action['where']);
 
         foreach ($versions as $version) {
             $this->routes[$version]->add($route);


Roquie:
Maybe, should I to give further explanation?

pafelin:
will this be merged?

Roquie:
Ping @jasonlewis

jasonlewis:
Sorry, slowly getting around to everything.

Do we need to check if the `$action` has the `where` key before using it?

Roquie:
Nope. https://github.com/dingo/api/blob/master/src/Routing/Router.php#L464

jasonlewis:
Of course! Thanks. :+1:

PR #811, opened 2016-01-12
Lumen 5.2 support

jeffberry:
Adds support for Lumen 5.2+

### DIFF
--- a/src/Provider/ApiServiceProvider.php
+++ b/src/Provider/ApiServiceProvider.php
@@ -65,7 +65,7 @@ abstract class ApiServiceProvider extends ServiceProvider
             'Dingo\Api\Console\Command\Docs',
         ]);
 
-        if (class_exists('Illuminate\Contracts\Foundation\Application', false)) {
+        if (class_exists('Illuminate\Foundation\Application', false)) {
             $this->commands([
                 'Dingo\Api\Console\Command\Cache',
                 'Dingo\Api\Console\Command\Routes',
--- a/src/Provider/ApiServiceProvider.php
+++ b/src/Provider/ApiServiceProvider.php
@@ -65,7 +65,7 @@ abstract class ApiServiceProvider extends ServiceProvider
             'Dingo\Api\Console\Command\Docs',
         ]);
 
-        if (class_exists('Illuminate\Foundation\Application', false)) {
+        if (class_exists('Illuminate\Contracts\Foundation\Application', false)) {
             $this->commands([
                 'Dingo\Api\Console\Command\Cache',
                 'Dingo\Api\Console\Command\Routes',
--- a/src/Console/Command/Cache.php
+++ b/src/Console/Command/Cache.php
@@ -79,7 +79,7 @@ class Cache extends Command
     /**
      * Get a fresh application instance.
      *
-     * @return \Illuminate\Foundation\Application
+     * @return \Illuminate\Contracts\Container\Container
      */
     protected function getFreshApplication()
     {
--- a/src/Http/Middleware/Request.php
+++ b/src/Http/Middleware/Request.php
@@ -5,11 +5,11 @@ namespace Dingo\Api\Http\Middleware;
 use Closure;
 use Exception;
 use Dingo\Api\Routing\Router;
+use Illuminate\Contracts\Container\Container;
 use Illuminate\Pipeline\Pipeline;
 use Dingo\Api\Http\RequestValidator;
 use Dingo\Api\Http\Request as HttpRequest;
 use Dingo\Api\Contract\Debug\ExceptionHandler;
-use Illuminate\Contracts\Foundation\Application;
 
 class Request
 {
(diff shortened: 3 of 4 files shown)


kylestev:
:+1:

jeffberry:
@jasonlewis any chance you can take a look at this soon? Would love to get this merged in so I can switch back to the original package & ship my code!

AriaEdo:
:+1:

ellisthedev:
Any word on this? We are starting to build out our API and want to use Lumen 5.2. We set everything up on 5.1 because of this bug; but since it's early we would love to update to 5.2 and keep rolling.

Cheers

lukeed:
+1 on lumen 5.2 now, may downgrade just so that I can continue development :cry:

ellisthedev:
Actually, looking at this Pull Request in more detail. This may not be the right way to upgrade to 5.2 support. I'm working through the upgrade guide of 5.1 to 5.2 and will open a new Pull Request with the right `Application` typehint.

jeffberry:
@ellisthedev the `Laravel\Lumen\Application` that you are going to attempt to use will most likely get you up and running. On Lumen. The reason this pull request uses the default Container contract is to keep this package working Laravel as well, which obviously does not use the `Laravel\Lumen\Application` class.

ellisthedev:
Derp, this is a cross-platform library. Forgot about that. :)

adamgoose:
:+1:

mattvb91:
:+1: @jasonlewis any chance getting this in? We are currently trying to upgrade to 5.2 and this seems to be the only issue left.

src/Provider/ApiServiceProvider.php
```diff
@@ -65,7 +65,7 @@ public function register()
             'Dingo\Api\Console\Command\Docs',
         ]);
 
-        if (class_exists('Illuminate\Foundation\Application', false)) {
+        if (class_exists('Illuminate\Contracts\Foundation\Application', false)) {
```
crynobone: Wouldn't this be wrong? 

we check for `Illuminate\Foundation\Application` because these commands only works on Laravel, not Lumen.

jeffberry: That's exactly why we're checking that the `Illuminate\Contracts\Foundation\Application` interface which [Laravel implements](https://github.com/laravel/framework/blob/5.2/src/Illuminate/Foundation/Application.php) exists and not for the `Illuminate\Contracts\Container\Container` interface which [Lumen solely implements](https://github.com/laravel/lumen-framework/blob/5.2/src/Application.php).

crynobone: `laravel/lumen` would load `illuminate/contracts`, correct? And `Illuminate\Contracts\Foundation\Application` are under `illuminate/contracts`, correct?

crynobone: You're checking `class_exists` here, and not `$this->app instanceof Application`.

jeffberry: @crynobone I'm not sure I follow you. `laravel/lumen` would load `illuminate/contracts` ? False. An application does not "load a namespace". Lumen implements the base contract `Illuminate\Contracts\Container\Container`. Laravel however, implements `Illuminate\Contracts\Foundation\Application`. That is why the author of this plugin is checking to see if that class exists (meaning Laravel is installed).

I'm not checking `class_exists`, the author of this plugin is checking `class_exists`. This is simply a (working) pull request to get the package fixed and working properly. You're only helping detract from the push this pull request needs to help myself and others get this package implemented!

crynobone: > I'm not sure I follow you. laravel/lumen would load illuminate/contracts ? False.

https://github.com/laravel/lumen-framework/blob/5.2/composer.json#L19-L38

illuminate/auth requires illuminate/contracts
illuminate/config requires illuminate/contracts
illuminate/container requires illuminate/contracts

How is it `false`? You can check `composer show -i` on a Lumen installation and see `illuminate/contracts` is install (or look inside your `vendor` folder.

> You're only helping detract from the push this pull request needs to help myself and others get this package implemented!

I'm helping not to introduce a **buggy** PR to be merged in.

crynobone: > That is why the author of this plugin is checking to see if that class exists (meaning Laravel is installed).

The author was correct to check for the concrete class `Illuminate\Foundation\Application` which means Laravel is installed. You're checking for `Illuminate\Contracts\Foundation\Application` which means `illuminate/contracts` is installed!.

jeffberry: You're correct, like I said I had a hard time following your poor explanation. I've reverted to the original line and would like to continue trying to get this merged in.

lasselehtinen:
Also running a Lumen 5.2. project which would very much like to use dingo/api :+1:

mattvb91:
Thanks @jasonlewis

PR #866, opened 2016-02-27
Change ModelNotFoundException to NotFoundHttpException

bkintanar:
This is needed since `ModelNotFoundException` throws error code 500. A change to `NotFoundHttpException` is needed in order for it to throw error code 404 instead.

ModelNotFoundException happens when using Laravel's explicit model binding and whenever the Id of the Model being queried doesn't exist.

### DIFF
--- a/src/Exception/Handler.php
+++ b/src/Exception/Handler.php
@@ -7,6 +7,8 @@ use ReflectionFunction;
 use Illuminate\Http\Response;
 use Dingo\Api\Contract\Debug\ExceptionHandler;
 use Dingo\Api\Contract\Debug\MessageBagErrors;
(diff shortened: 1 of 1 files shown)


mrgrain:
Strongly against this, as it dramatically changes the semantics. A `ModelNotFoundException` **could** mean you want to throw a `NotFoundHttpException`, but you might also handle it differently.

Also, it is easy enough to implement yourself on application level.

tymondesigns:
:-1: 100% agree with @mrgrain

The fact that a model is not found has nothing at all to do with http

jasonlewis:
Correct, handle the exceptions how you want within your own application.

PR #867, opened 2016-02-29
Filter availbles includes from transformer when eager loads

gffuma:


jasonlewis:
See bf29beb501613d0eee342ac3e962113debddbce.

Can you provide my a reason as to why we should be intersecting the available includes? With it, because of how Fractal does the nested includes, it would strip out things like `nested.include` which would result in lazy loading a bunch of things.

gffuma:
Ok, i understand the problem of not eager load the nested includes, but how can i deal with situation where the relation of load is invalid, actually this cause exception like:
`Call to undefined method Illuminate\\Database\\Query\\Builder::x()`

Do you think is possible to reimplement this part in a smarter way, to check if given relation to load is valid?
https://github.com/dingo/api/blob/master/src/Transformer/Adapter/Fractal.php#L92

jasonlewis:
Why is it invalid?

I'll try playing around with this a bit more later. See if I can come up with a better solution.

tembra:
+1
I got the same problem.

@jasonlewis What @gffuma wants to say is once the guy that is consuming the API can just parse any include that he wants on query paremeter (e.g `/api/posts?include=xxx`), this will cause a `BadMethodCallException` error since `xxx` is not an available include (I mean `xxx` method doesn't exists on `Post` model).

I didn't want to change the package itself, so I manage the problem in `Exceptions/Handler` class since the exception just throw the first block of try/catch that is too nested inside. This is my code to manage this problem for while:
```php
public function report(Exception $e)
{
    $report = true;
    $stackTrace = $e->getTrace();
    foreach ($stackTrace as $trace) {
        if ($trace['function'] == 'load' && substr($trace['file'], -11) == 'Fractal.php') {
            app('Dingo\Api\Http\Response\Factory')->errorBadRequest('The include requested is invalid!');
            $report = false;
            exit;
        }
    }
    if ($report) {
        parent::report($e);
    }
}
```

jasonlewis:
This is still on my to-do of things to look at to improve.

specialtactics:
I believe this has been fixed in a PR Merged very recently, so closing.

PR #890, opened 2016-03-12
Routes List Command for Lumen

tembra:
Created `RoutesLumen.php` file to provide `api:routes` command for Lumen. The file was created using the existing `Routes.php` for Laravel’s `api:routes` command and was only injected the single dependency `$options` array from Laravel's `route:list` command.

Modified the `ApiServiceProvider.php` to register the new `RoutesLumen.php`.

Modified `getActionName()` method from `Route.php` file to find `uses` index instead of `controller` on `action` property array. This correction is needed because when you define a named route using
the `as` array key, the `controller` property remains empty.


tembra:
Will make some corrections and send again later.

PR #892, opened 2016-03-14
Pass server variables to internal request

mcharytoniuk:
In my API implementation I am trying to return some URLs generated with Laravel's route() helper. The problem is that Dispatcher does not pass server variables from actual request into the internal request and links generated with route() are quite random (pointing to http://localhost for example).

Is this ok to pass all actual server variables instead of the empty array? It would solve my problem.

### DIFF
--- a/src/Dispatcher.php
+++ b/src/Dispatcher.php
@@ -466,7 +466,15 @@ class Dispatcher
             $uri = rtrim($this->getRootRequest()->root(), '/').'/'.ltrim($uri, '/');
         }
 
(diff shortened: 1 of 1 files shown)


jasonlewis:
I can't see it being an issue.

jasonlewis:
We'll wait and see what happens though. Cheers. :+1:

mcharytoniuk:
It shines when you are trying to use server variables or things like `route()` at internal request level. It solves my issue. Thanks. :)

PR #927, opened 2016-03-22
Add support for JsonResponse in exception handler

behzadsh:
If the registered custom handler, return response with type of `JsonResponse`, it should return the response and not passed to newly generated `Response` class.

It would be better to check if the returned response from custom handler is instance of `Symfony\Component\HttpFoundation\Response` rather than `Illuminate\Http\Response` and `Illuminate\Http\JsonResponse`. But since the package don't depend on `symfony/http-foundation` I did not change it to symfony request class.

### DIFF
--- a/src/Exception/Handler.php
+++ b/src/Exception/Handler.php
@@ -4,6 +4,7 @@ namespace Dingo\Api\Exception;
 
 use Exception;
 use ReflectionFunction;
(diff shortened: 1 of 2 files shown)


hskrasek:
Handled by 84e9446abcf162810699c8dca2519c55ec6400ed

PR #935, opened 2016-03-24
Update Http\Response\Factory.php

Grisou13:
Add content to a created request. Even though the HTTP 201 is not intended to send data back, sending some sort of content, like an id of the created resource to the requester can be very usefull.

### DIFF
--- a/src/Http/Response/Factory.php
+++ b/src/Http/Response/Factory.php
@@ -39,9 +39,9 @@ class Factory
      *
      * @return \Dingo\Api\Http\Response
      */
-    public function created($location = null)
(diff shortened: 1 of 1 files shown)


Ryuske:
Refer to #942 for more information regarding why this change should be made.

I don't know if this way (non-breaking) or the way I did it which maintains compatibility between responses (but causes breaking changes) is better.

jasonlewis:
I'd honestly prefer to keep this simple. If you need to use `created` with an `item` or `collection` just chain a call to `setStatusCode` afterwards.

PR #942, opened 2016-03-25
Adding ability to return content with http factory created - maintain compatability with ->item()

Ryuske:
As per http://jsonapi.org/format/#crud-creating, the 201 Created status should response with a document containing the primary resource created.

Otherwise you should return 204 No Content.

This fork maintains compatibility with item() which seems desired to be when returning the primary resource. However, maybe for non-breaking changes, the PR here (#935) would be better?


jasonlewis:
I'm not going to assume everyone is using transformers.

PR #945, opened 2016-03-28
Fixed getIterableRoutes not property aggregating static routes

Tschitsch:


jasonlewis:
I've gone ahead and pushed a little more robust fix for this.

PR #950, opened 2016-03-30
Allow custom limiter when Dingo\Api\Contract\Http\RateLimit\Throttle also implements HasRateLimiter.

crynobone:
This would allow custom Throttle to define the rate limiter key resolver within the class (instead of declaring it globally). This would allow different Throttle to use different rate limiter key values.

Signed-off-by: crynobone <crynobone@gmail.com>

### DIFF
--- a/src/Contract/Http/RateLimit/HasRateLimiter.php
+++ b/src/Contract/Http/RateLimit/HasRateLimiter.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Dingo\Api\Contract\Http\RateLimit;
+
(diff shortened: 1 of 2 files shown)


crynobone:
Travis-CI failing on an unrelated issues:

```
 [RuntimeException]                                                           
  Could not load package shopware/shopware in http://packagist.org: [Unexpect  
  edValueException] Could not parse version constraint ~v2.0.0-rc1: Invalid v  
  ersion string "~v2.0.0-rc1"
```

PR #961, opened 2016-04-03
Use specific Laravel version

vinkla:
We wont know if this package will break in Laravel 5.3. Laravel doesn't follow SemVer. This way we can get ahead of future problems.


vinkla:
StyleCI seems to fail. Though, it has nothing to do with this pull request since it only updates the `composer.json` file.

jasonlewis:
Conflicts due to other PR, if you can't resolve I'll just merge it locally when I get a chance.

vinkla:
@jasonlewis I've replaced it with #969

PR #962, opened 2016-04-03
Update phpunit config

vinkla:
We should not allow risky tests and warnings.


vinkla:
StyleCI seems to fail. It has nothing to do with this PR since it only updates the `phpunit.xml.dist` file.

PR #963, opened 2016-04-03
Test against php 7.0

vinkla:
We should/could start testing against PHP 7.0.


vinkla:
StyleCI seems to fail. It has nothing to do with this PR since it doesn't have any syntax changes.

PR #984, opened 2016-04-12
Add api route helper function

TheGIBSON:


TheGIBSON:
ok, version('v1')->route() does the same, so this isn't neccessary. closing

Issue #678, opened 2015-10-07
Internal Requests and Named Routes


brunogaspar:
I was trying to do an internal request with a named route but it seems a though job.

The only way, which it's very ugly, i could find to make it work, is by having something like the following:

```php
$this->api->get(app('api.url')->route('customer.show', $id));
```

But something like:

```php
$this->api->route('customer.show', $id);
# and/or
$this->api->version('V3')->route('customer.show', $id);
```

Would be cool to have available on the dispatcher which also looks more cleaner.

> Sorry if this was asked before, but i couldn't find any issue with this "problem".

jasonlewis:
This did initially work, but I took it out, and for some reason that I can't remember it right now. I'll have another look when I can (probably next week), but I believe it was something to do with the different routing adapters.

brunogaspar:
Thanks @jasonlewis there's no rush, i'm using the "ugly" way for now, works fine =)

Luddinus:
+1

dpgover:
+1

guiwoda:
:+1:

dpgover:
I was trying to implement this workaround and found some problems:

First, this `$this->api->get(app('api.url')->route('customer.show', $id));` did not work. I had to set the version of the API first using the `version` method on the urlGenerator (and yes, I have the default version configured in the config file).

Second issue is that my API works with http and https. If I set a domain like http://myapi.foo.bar, the https urls do not work. I tried using the prefix like `/`, it all worked fine, until I tried to make an internal request. Those requests fail if the prefix is `/`.

Hopes this helps fixing the issue, seems that this is all about the same problem.

Cheers,
And thanks for the awesome awesome awesome package.

dpgover:
Just found out that if the configured domain has a trailing slash, all requests throw an HttpNotFoundException...

agentphoenix:
Ran into this as well and would love to have a simpler way to reference named routes.

PR #988, opened 2016-04-13
Fix Disptacher "createRequest" method to use named routes

TheGIBSON:
Now we can use:

``` PHP
app('Dingo\Api\Dispatcher')->get(version('v1')->route('test'));
```

This also fixes issue: #678

### DIFF
--- a/src/Dispatcher.php
+++ b/src/Dispatcher.php
@@ -458,11 +458,11 @@ class Dispatcher
     {
         $parameters = array_merge($this->parameters, (array) $parameters);
 
-        $uri = $this->addPrefixToUri($uri);
-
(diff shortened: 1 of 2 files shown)


src/Dispatcher.php
```diff
@@ -539,7 +539,7 @@ protected function dispatch(InternalRequest $request)
 
             $response = $this->router->dispatch($request);
 
-            if (! $response->isSuccessful()) {
+            if (! $response->isSuccessful() && ! $response->isEmpty()) {
```
TheGIBSON: Status code "304" must be valid. e.g. when response is cached image.

Issue #964, opened 2016-04-03
Internal routing bug: doesn't work with middleware


assembledadam:
Any internal request in conjunction with a route that contains middleware, like this:

```
$api->version('v1', ['middleware' => 'test'], function ($api) {

});
```

Results in a:

```
[ReflectionException]      
  Class test does not exist 
```

Despite it being define properly inside Kernel.php:

```
 protected $routeMiddleware = [
        'test' => \App\Http\Middleware\Test::class,
        'auth' => \App\Http\Middleware\Authenticate::class,
        'auth.basic' => \Illuminate\Auth\Middleware\AuthenticateWithBasicAuth::class,
        'guest' => \App\Http\Middleware\RedirectIfAuthenticated::class,
        'throttle' => \Illuminate\Routing\Middleware\ThrottleRequests::class,
    ];
```

I've created a test case to illustrate the bug here https://github.com/assembledadam/laravel

Just run `php artisan testcase` in the root folder, and to view the middleware working visit `localhost/index.php/test/blahblah`

Stack trace (from my application, not the test case!):

```
n Error Occured:
Class oauth does not exist
/usr/www/app/vendor/laravel/framework/src/Illuminate/Container/Container.php
Line: 738
#0 /usr/www/app/vendor/laravel/framework/src/Illuminate/Container/Container.php(738): ReflectionClass->__construct('oauth')
#1 /usr/www/app/vendor/laravel/framework/src/Illuminate/Container/Container.php(633): Illuminate\Container\Container->build('oauth', Array)
#2 /usr/www/app/vendor/laravel/framework/src/Illuminate/Foundation/Application.php(697): Illuminate\Container\Container->make('oauth', Array)
#3 /usr/www/app/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(123): Illuminate\Foundation\Application->make('oauth')
#4 [internal function]: Illuminate\Pipeline\Pipeline->Illuminate\Pipeline\{closure}(Object(Dingo\Api\Http\InternalRequest))
#5 /usr/www/app/vendor/laravel/framework/src/Illuminate/Routing/Pipeline.php(32): call_user_func(Object(Closure), Object(Dingo\Api\Http\InternalRequest))
#6 /usr/www/app/vendor/dingo/api/src/Http/Middleware/PrepareController.php(45): Illuminate\Routing\Pipeline->Illuminate\Routing\{closure}(Object(Dingo\Api\Http\InternalRequest))
[...]

PR #1007, opened 2016-04-28
Slightly hacky fix to address stale controllers when firing internal requests + InternalRequests bugfix

assembledadam:
See https://github.com/dingo/api/issues/964#issuecomment-208845502


thilanga:
No activities.

PR #1011, opened 2016-04-28
Throttle groups

### LINKED ISSUE #1010: Misunderstanding about custom throttles
Hello,
after several hours of tracing the source code, I finally realise the use of custom throttles.

It is meant to determine if a route is to be throttled or not by searching through registered throttles by matching a certain condition (defined by the custom throttle), and it **IS NOT route-specific**.

`['middleware' => 'api.throttle', 'limit' => 100, 'expires' => 5]` is route-specific.

Possible use: Check if User is Paid user or not, then apply the appropriate  'limit' and 'expires' value.

For Lumen: `bootstrap/app.php`

``` php
app('Dingo\Api\Http\RateLimit\Handler')->extend(new PaidUserThrottle(['limit' => 1000, 'expires' => 10]));
app('Dingo\Api\Http\RateLimit\Handler')->extend(new GeneralUserThrottle(['limit' => 200, 'expires' => 10]));
```

And in `routes.php`:

``` php
    $api->group(['middleware' => 'api.throttle'], function ($api) {
        $api->post('auth', 'App\Http\Controllers\Api\V1\AuthController@login');            // Authenticate
    });
```

(do not include "'limit' => 100, 'expires' => 5" in the array)

My point is, it will be nice to mention what custom throttles really do and show a route example in the docs...

NicksonYap:
Allow setting $keyPrefix so that custom throttles can be grouped conditionally.
Trying to resolve #1010 

``` php
<?php

use Illuminate\Container\Container;
use Dingo\Api\Http\RateLimit\Throttle\Throttle;

class ExampleThrottle extends Throttle
{

    /**
     * Example throttle will be matched unconditionally.
     *
     * @param \Illuminate\Container\Container $container
     *
     * @return bool
     */
    public function match(Container $app)
    {
        // Perform some logic here and return either true or false depending on whether
        // your conditions matched for the throttle.

        return true;
    }

    /**
     * Define throttle groups conditonally.
     *
     * @return void
     */
    public function group(){
        return function ($app, $request) {

            //return null to apply thottle globally
            return md5($request->path());
        };
    }
}
```

### DIFF
--- a/src/Http/RateLimit/Handler.php
+++ b/src/Http/RateLimit/Handler.php
@@ -330,7 +330,7 @@ class Handler
 
         $this->throttles->push($throttle);
     }
-    
+
     /**
      * Get the key prefix.
      *
@@ -339,10 +339,8 @@ class Handler
     public function getKeyPrefix()
     {
         if (method_exists($this->throttle, 'group') && $this->throttle->group()) {
-            
             return call_user_func($this->throttle->group(), $this->container, $this->request);
         } else {
-            
             return md5($this->request->path());
         }
     }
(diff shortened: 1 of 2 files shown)


thilanga:
No activities

PR #1026, opened 2016-05-12
Add extra conditions array to Basic auth provider

prapats:
As the method `onceBasic` also take the `extraConditions` array as a parameter (https://github.com/laravel/framework/blob/5.2/src/Illuminate/Auth/SessionGuard.php#L289), I made the `Basic` provider be able to take the array too.


specialtactics:
Hi @prapats 

Looks good, can you just re-run styleCI and fix whatever issue that was? Then I can merge

PR #1043, opened 2016-05-29
Use the bootstrapPath method if it exists

agentphoenix:
Laravel 5.2.23 introduced the bootstrapPath method in the Application container. In the event someone is using non-default paths (such as the bootstrap path), there’s no way to compile API routes using the built-in console job. This pull request checks to see if the bootstrapPath method exists and uses it if it does.


specialtactics:
Merged !

Issue #981, opened 2016-04-11
array_merge(): Argument #1 is not an array


sagaio:
Current build: `Laravel v5.1.33 / Dingo/API dev-master 7fa939c`
After an update and then a rollback(received 404, as some other issues present), I receive this error:
```
'array_merge(): Argument #1 is not an array' in C:\project\vendor\dingo\api\src\Routing\Route.php:205
```
I'm not sure what is going on but this happens when I'm trying to login to my app, therefor moving from a route that is not using the middleware `auth` to a route that does(from Laravel route to API route).

I did a workaround in `Route.php` but I hope someone can explain to me what is going wrong...
Instead of:
```
$this->middleware = array_merge($this->middleware, $controllerMiddleware);
```
I added this and it's now working...
```
if(count($controllerMiddleware)) {
    $this->middleware = array_merge($this->middleware, $controllerMiddleware);
}
```

HtmHell:
+1 it happens when I use a middleware in the regular Laravel's route:
```
Route::group(['namespace' => 'Admin', 'middleware' => 'admin:guest'], function() use($p)
{...
```

hskrasek:
@sagaio Was this in Laravel or Lumen?

hskrasek:
I can't re-create this in the most recent tagged version of the package. I went over your pull request, and it's not really fixing the underlying issue, it's just masking it. The underlying issue is due to the adapter returning middleware as a string (or something) that isn't an array.

When I pull this into a fresh install of Laravel 5.2 though, I can add a middleware to a non-dingo route and see it correctly returning an array of middleware.

I have a fix that can ensure the issue doesn't resurface though.

PR #1073, opened 2016-06-19
Fixed a small issue with middlewares merge issue

meSingh:
A small middlewares merge fix as suggested here:
https://github.com/dingo/api/issues/981#ref-commit-0269fee


hskrasek:
As discussed in the linked issue, the PR doesn't fix the underlying issue but hides it. I've applied a fix that will remove the issue though

PR #1083, opened 2016-06-23
Code clean up for `register` method in src/Transformer/Factory.php

thilanga:
I did clean the code bit as strong typing on this method makes the method more neat and clean.
Also modified testRegisterParameterOrder to reflect new changes.


hskrasek:
Current method signature is intended, to allow for cleaner usage.

PR #1104, opened 2016-07-06
Update Helpers.php

kevinkhill:
Simple method for using the UrlGenerator in a controller.


kevinkhill:
I am now realizing this doesn't handle route parameters... I will submit another.

PR #1105, opened 2016-07-06
Use stricter tests assertions

lucasmichot:
Use stricter tests assertions


lucasmichot:
I do think there should be a config key for Styleci, that enable the style to be checked only on added or modified files within a PR
Ping @GrahamCampbell

GrahamCampbell:
Nah, StyleCI isn't built for that. TBH, the PR status is a waste of time IMO, but people seem to want it.

GrahamCampbell:
The real use case is turn it on, and forget about it. I have no interest in partial fixing.

hskrasek:
Looks good to me @lucasmichot

PR #1136, opened 2016-07-28
rate limit: Limit should apply to all routes

jwdeitch:
Exceeding the rate limit should return 403 for all further requests; the route of the request should not matter.

The current behavior is such that exceeding the rate limit will only 403 that particular route.

### DIFF
--- a/src/Http/RateLimit/Handler.php
+++ b/src/Http/RateLimit/Handler.php
@@ -103,8 +103,6 @@ class Handler
         } elseif ($limit > 0 || $expires > 0) {
             $this->throttle = new Route(['limit' => $limit, 'expires' => $expires]);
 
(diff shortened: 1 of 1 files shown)


thilanga:
@jwdeitch This is a good PR, Could you resolve the conflict and update the PR?

specialtactics:
Wouldn't this depend on whether the api throttle is placed in a route group or a particular route? If particular route or even group within larger group - then it shouldn't I would think apply to *all* routes?

johannesschobel:
Actually, the middleware should throw a `http status 429 (too many requests)` as indicated here ( https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429 )

Issue #319, opened 2015-01-04
HTTP 429 when rate limits have been exceeded


stelgenhof:
Currently a 403 HTTP code is issued when rate limits are exceeded. As per RFC http://tools.ietf.org/html/rfc6585, would 429 not be a better status code?

Andreyco:
+1

jasonlewis:
I'm not the best at deciding on status codes. I've used non-HTTP spec status codes in this package already so probably wouldn't hurt. I believe I went with 403 as that's what GitHub and many others are using for exceeded rate limits.

Send a PR though if you like.

stelgenhof:
403 just means forbidden, whereas 429 means "Too Many Requests" which in my opinion serves much better the situation when rates have been exceeded. I agree it's a grey area :) By the way others like Twitter are using 429 as an error code for example.

Would it be an idea to make it configurable then?

Andreyco:
No configuration please.

429 is more descriptive than 403 so lets settle with 429 only. I will take a look and send a PR.

PR #1158, opened 2016-08-17
Updated RateLimit exception response to RFC6585 specifications

dlnandrsn:
> The 429 status code indicates that the user has sent too many requests in a given amount of time ("rate limiting").
> 
> The response representations SHOULD include details explaining the condition, and MAY include a Retry-After header indicating how long to wait before making a new request.


dlnandrsn:
See #319

src/Http/Middleware/RateLimit.php
```diff
@@ -64,7 +64,8 @@ public function handle($request, Closure $next)
         $this->handler->rateLimitRequest($request, $route->getRateLimit(), $route->getRateLimitExpiration());
 
         if ($this->handler->exceededRateLimit()) {
-            throw new HttpException(403, 'You have exceeded your rate limit.', null, $this->getHeaders());
+            $retryAfter = $this->getHeaders()['X-RateLimit-Reset'] - time();
+            throw new TooManyRequestsHttpException($retryAfter, 'You have exceeded your rate limit.');
```
lucasmichot: I would just change the status code to 429 only, but keep all of the headers:
```php
if ($this->handler->exceededRateLimit()) {
    $headers = $this->getHeaders();
    $headers['Retry-After'] = $headers['X-RateLimit-Reset'] - time();

    throw new HttpException(429, 'You have exceeded your rate limit.', null, $headers);
}
```

hskrasek: I agree with @lucasmichot. Would be nice if we could throw the actual named exception, but seems Symfony doesn't want us to be able to set custom headers on that exception.

dlnandrsn: @hskrasek & @lucasmichot - The reason i wanted to specifically throw the correct named exception was so it could be caught in custom exception responses. It is not really feasible to catch all thrown HttpExceptions (at least not in my use case) and modify them, i guess..

Is it maybe possible to create a custom HTTP exception that implements HttpExceptionInterface so others are able to hook into it and define a custom response? What are everyones thoughts?

lucasmichot: Well you totally tweak exceptions the way you want...
If you are using Laravel, in `\App\Exceptions\Handler` you can do something like:

```php
/**
 * {@inheritdoc}
 */
protected function prepareException(Exception $e)
{
    $e = parent::prepareException($e);

    if ($e instanceof \Symfony\Component\HttpKernel\Exception\HttpException) {
        if ($e->getStatusCode() === 403 && $e->getMessage() === 'You have exceeded your rate limit.') {
            $headers = $e->getHeaders();
            $retryAfter = $headers['X-RateLimit-Reset'] - time();

            $e = new \Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException(($retryAfter, 'You have exceeded your rate limit.');
        }
    }

    return $e;
}
```

Then you can catch you `TooManyRequestsHttpException`...

Is that what you meant?

src/Exception/RateLimitExceededException.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Dingo\Api\Exception;
+
+use Exception;
+use Symfony\Component\HttpKernel\Exception\HttpException;
+
+class RateLimitExceededException extends HttpException
+{
+    /**
+     * Create a new unknown version exception instance.
```
lucasmichot: ?

dlnandrsn:
@hskrasek & @lucasmichot - Alright, i have created a custom exception handler for rate limit exceeded exceptions. This now keeps the custom X-* headers, still implements the "Retry-After" header and 429 status, and also allows users to easily hook into the exception to define a custom response.

Sorry about the earlier commit discarding the custom headers. I hope this solution is acceptable  (:

src/Http/Middleware/RateLimit.php
```diff
@@ -64,8 +64,9 @@ public function handle($request, Closure $next)
         $this->handler->rateLimitRequest($request, $route->getRateLimit(), $route->getRateLimitExpiration());
 
         if ($this->handler->exceededRateLimit()) {
-            $retryAfter = $this->getHeaders()['X-RateLimit-Reset'] - time();
-            throw new TooManyRequestsHttpException($retryAfter, 'You have exceeded your rate limit.');
+            $headers = $this->getHeaders();
+            $headers['Retry-After'] = $headers['X-RateLimit-Reset'] - time();
```
lucasmichot: This should be calculated within the exception then

src/Http/Middleware/RateLimit.php
```diff
@@ -64,7 +64,9 @@ public function handle($request, Closure $next)
         $this->handler->rateLimitRequest($request, $route->getRateLimit(), $route->getRateLimitExpiration());
 
         if ($this->handler->exceededRateLimit()) {
-            throw new HttpException(403, 'You have exceeded your rate limit.', null, $this->getHeaders());
+            $headers = $this->getHeaders();
+            $headers['Retry-After'] = $headers['X-RateLimit-Reset'] - time();
```
lucasmichot: This should be calculated within the exception

tests/Http/Middleware/RateLimitTest.php
```diff
@@ -123,11 +123,12 @@ public function testRateLimitingFailsAndHeadersAreSetOnException()
                 return new Response('foo');
             });
         } catch (HttpException $exception) {
-            $this->assertSame(403, $exception->getStatusCode());
+            $this->assertSame(429, $exception->getStatusCode());
             $this->assertSame('You have exceeded your rate limit.', $exception->getMessage());
             $this->assertArrayHasKey('X-RateLimit-Limit', $exception->getHeaders());
             $this->assertArrayHasKey('X-RateLimit-Remaining', $exception->getHeaders());
             $this->assertArrayHasKey('X-RateLimit-Reset', $exception->getHeaders());
+            $this->assertArrayHasKey('Retry-After', $exception->getHeaders());
```
lucasmichot: An assertion should be done on the value of `Retry-After`

src/Http/Middleware/RateLimit.php
dlnandrsn: Ok fair enough, ill fix that now.

tests/Http/Middleware/RateLimitTest.php
dlnandrsn: @lucasmichot - Before i commit the assertion change, do you have an opinion on how i should perform the Retry-After assertion? Is the below snippet fine?

```php
try {
    $this->middleware->handle($request, function ($request) {
        return new Response('foo');
    });
} catch (HttpException $exception) {
    $this->assertSame(429, $exception->getStatusCode());
    $this->assertSame('You have exceeded your rate limit.', $exception->getMessage());
    $this->assertSame($exception->getHeaders()['X-RateLimit-Reset'] - time(), $exception->getHeaders()['Retry-After']);
    $this->assertArrayHasKey('X-RateLimit-Limit', $exception->getHeaders());
    $this->assertArrayHasKey('X-RateLimit-Remaining', $exception->getHeaders());
    $this->assertArrayHasKey('X-RateLimit-Reset', $exception->getHeaders());
}
```

lucasmichot:
can you squash this pr maybe?

hskrasek:
@lucasmichot I don't have the power to, but I can ask @jasonlewis to enable squash and merge. That way this PR can be squashed when we merge it in.

Also, thanks for the review on this

src/Exception/RateLimitExceededException.php
```diff
@@ -0,0 +1,25 @@
...
+     * @param string     $message
+     * @param \Exception $previous
+     * @param array      $headers
+     * @param int        $code
+     *
+     * @return void
+     */
+    public function __construct($message = null, Exception $previous = null, $headers = [], $code = 0)
+    {
+        $this->headers = array_key_exists('X-RateLimit-Reset', $headers) ? $headers['Retry-After'] = $headers['X-RateLimit-Reset'] - time() : $headers;
```
lucasmichot: `\Symfony\Component\HttpKernel\Exception\HttpException::$headers`is private !

You should do this instead:
```php
if (array_key_exists('X-RateLimit-Reset', $headers)) {
    $headers['Retry-After'] = $headers['X-RateLimit-Reset'] - time();
}

parent::__construct(429, $message ?: 'You have exceeded your rate limit.', $previous, $headers, $code);
```

hskrasek:
@dlnandrsn If you can make the change @lucasmichot suggested, I'll be able to merge this.

PR #1174, opened 2016-08-27
Revert "Remove useless import."

hskrasek:
Reverts dingo/api#1099


lucasmichot:
This branch should probably be deleted @hskrasek

PR #1009, opened 2016-04-28
Use ::class notation to reference to classes

driesvints:
As promised, replace all hardcoded references to classes with the `::class` notation. This allows for cleaner code and better references to used classes.

It's best that you review this a bit to see if I didn't miss anything or replaced the wrong class since a lot of classes have the same naming as Illuminate classes.

PR #1206, opened 2016-09-22
Use ::class notation to reference to classes

lucasmichot:
Use `::class` notation to reference to classes

Integrates original commits  https://github.com/dingo/api/pull/1009 and replace it


driesvints:
@lucasmichot thank you for picking this up :)

PR #1270, opened 2016-12-10
API_STRICT only for specific URLs

Barmunksu:
API_STRICT only for specific URLs with $request->is pattern

### DIFF
--- a/config/api.php
+++ b/config/api.php
@@ -114,6 +114,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Strict Mode only for specific URLs
+    |--------------------------------------------------------------------------
+    |
+    | Enabling strict mode for specific URLs only if API_STRICT=true.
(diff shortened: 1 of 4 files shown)


thilanga:
@Barmunksu Can you explain me what is this really useful or your use case?

thilanga:
No Response from the author

lloricode:
@thilanga 

hi this is helpful when you need to make some listener for external sites, like verifying user from url on their email, redirecting this will use `GET` with no headers, so this api listener will cannot interacting with given url if it has a `strict` rule to all routes

lloricode:
please merge this or, enhance if needed

specialtactics:
Can't you just not have the struct rule on all routes? @lloricode

lloricode:
yes but it is helpful when uses `strict`, the only downside is cannot control which route can be apply for this

specialtactics:
I understand @lloricode unfortunately this cannot be a priority right now to look at.

PR #1314, opened 2017-01-24
Laravel 5.4 Illuminate/Routing/Route class changes.

denpamusic:
getUri() method was renamed to uri(), and getMethods() method was
renamed to methods().
See https://laravel.com/api/5.4/Illuminate/Routing/Route.html

### DIFF
--- a/src/Routing/Adapter/Laravel.php
+++ b/src/Routing/Adapter/Laravel.php
@@ -5,8 +5,8 @@ namespace Dingo\Api\Routing\Adapter;
 use Illuminate\Http\Request;
 use Illuminate\Routing\Route;
 use Illuminate\Routing\Router;
-use Illuminate\Routing\RouteCollection;
 use Dingo\Api\Contract\Routing\Adapter;
+use Illuminate\Routing\RouteCollection;
 use Dingo\Api\Exception\UnknownVersionException;
 
 class Laravel implements Adapter
--- a/src/Routing/Adapter/Laravel.php
+++ b/src/Routing/Adapter/Laravel.php
@@ -119,7 +119,7 @@ class Laravel implements Adapter
      */
     public function getRouteProperties($route, Request $request)
     {
-        return [$route->getUri(), $route->getMethods(), $route->getAction()];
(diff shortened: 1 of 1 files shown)


denpamusic:
Sorry, I'll add method_exist check.

andrewmclagan:
any chance of a merge on this anytime soon?

src/Routing/Adapter/Laravel.php
```diff
@@ -5,8 +5,8 @@
 use Illuminate\Http\Request;
 use Illuminate\Routing\Route;
 use Illuminate\Routing\Router;
-use Illuminate\Routing\RouteCollection;
 use Dingo\Api\Contract\Routing\Adapter;
+use Illuminate\Routing\RouteCollection;
```
bkintanar: Is this change necessary? I think we can revert this.

denpamusic: No, it's not. It actually make no sense for me, but Style CI complains about it https://styleci.io/analyses/qoQRrp

andrewmclagan: Agree, should follow the code style? ;-)

bkintanar: Well that's weird.. @GrahamCampbell any input on this weird Style CI analysis?

andrewmclagan: Any chance to merge this PR with the suggested style-ci changes?

jasonlewis:
5.4 has been tagged stable for less than 24 hours. In no way should you expect this package to work out of the box with it.

src/Routing/Adapter/Laravel.php
crynobone: > Is this change necessary? I think we can revert this.

Same length, it would then order the same length by alphanumeric order, so `D` before `I`.

denpamusic:
Of course not. I mostly shared this PR to save time for developers like myself who tried updating dev environment to see what breaks with 5.4.

jasonlewis:
Jolly good!

PR #1316, opened 2017-01-25
L5.4: Rename the router's middleware method to aliasMiddleware

rhwilr:
In Laravel 5.4 the routers `middleware` Method was renamed to `aliasMiddleware` see: https://github.com/laravel/framework/commit/06303d924b645dcde1913c79780ab63423426f29

In addition to this I updated the composer version constraints to allow illuminate 5.4 components to be installed which should make this package Laravel 5.4 compatible. I could not find any other issues in my testing.

Also this PR: https://github.com/dingo/blueprint/pull/54


src/Provider/LaravelServiceProvider.php
```diff
@@ -119,6 +119,27 @@ protected function addRequestMiddlewareToBeginning(Kernel $kernel)
     }
 
     /**
+     * Register a short-hand name for a middleware. For Compatability
+     * with Laravel < 5.4 check if aliasMiddleware exists since this
+     * method has been renamed.
+     *
+     * @param $name
```
lucasmichot:
```
@param string $name
@param string $class
  
@return \Illuminate\Routing\Router
```

rhwilr: thx, I added the typehints

PR #1354, opened 2017-03-12
JSON optional formatting feature

vinterskogen:
Hello.

Thanks, for the great API building package! 

This PR contains changes to JSON formatter, that provides optional formatting for JSON output.

## The problem

I discovered, that **default JSON formatter in Dingo does not provide a configurable way to tweak JSON-encoded output**. 

This issue makes sense, because the developers, that  build APIs using Dingo, (with essential chance) would like to configure their APIs' JSON output (to satisfy a various requirements for its formatting and structure) in certain ways, such as (options for `json_encode` PHP function):

- JSON_PRETTY_PRINT
- JSON_PRESERVE_ZERO_FRACTION
- JSON_NUMERIC_CHECK
- JSON_UNESCAPED_SLASHES
- JSON_FORCE_OBJECT
- ... 

And others. See: http://php.net/manual/ru/function.json-encode.php

This is really a very common situation, but for now that can't be done without implementing (by each developer) an own custom formatter, that handles the necessary JSON formatting operations. 

Not a good practice, as I conclude. Especially when realize, that the final goal of this package, is to get rid of re-implementing any extra-common API-specific functionality over again.

## The solution 

To solve the specified problem I **added an optional and configurable functionality to Dingo's default JSON formatter, that aims to deal with the requirements to JSON output formatting and structure**, and which can be conveniently configured by the APIs developers. 

Changes in this PR are focused on implementing configurable JSON encoding options feature and adds support for the JSON pretty print option.

It's quite important, because in general _API is a developer's interface to the application_, so it has to be comfy to interact with it (even when using pure CLI tools). The Github's API stands for this idea too - check its pretty printed output:

```
curl https://api.github.com/repos/dingo/api
```

JSON pretty print feature in this PR comes with various indent styles support (tab or spaces) and configurable indent size (the number of chars, that are used to indent one level). 

JSON optional formatting code is concluded in `JsonOptionalFormatting` trait, that is used by `Json` formatting class. It is a minimal functioning solution and, by design, it can be easily extended by any number of other `json_encode()` options in the future. 

## The Documentation

The following paragraphs, I guess, can be included to the [Responses](https://github.com/dingo/api/wiki/Responses) page on Wiki.

### Formats' options

**JSON pretty print**

JSON and JSONP formats provides a pretty print option, that beautifies the output of your API and makes it more comfortable to read by mortals. Pretty print feature comes with various indent styles support (tab or spaces) and configurable indent size.

By default, pretty print option is turned off. To enable and configure it you should place the following options to `.env` file:

- `API_JSON_FORMAT_PRETTY_PRINT_ENABLED` - switches JSON pretty print option (boolean)
- `API_JSON_FORMAT_INDENT_STYLE` - sets the indent style: 'tab' or 'spaces' (string)
- `API_JSON_FORMAT_INDENT_SIZE` - sets the number of chars, that are used to indent one level for 'space' indent style (integer, >= 1)

> **Note:** Behavior for options `API_JSON_FORMAT_INDENT_STYLE` and `API_JSON_FORMAT_INDENT_SIZE` is inspired by the same idea, that is behind the [EditorConfig](http://editorconfig.org/) options with the same titles. It's good for you to be familiar with that.

Be sure, that options are loading in `config/api.php` file (in Response Formats section). The default `formatsOptions` array look like this:

```
'formatsOptions' => [
  'json' => [
    'pretty_print' => env('API_JSON_FORMAT_PRETTY_PRINT_ENABLED', false),
    'indent_style' => env('API_JSON_FORMAT_INDENT_STYLE', 'space'),
    'indent_size'  => env('API_JSON_FORMAT_INDENT_SIZE', 2),
  ],
],
```

## Tests

Test classes for `Json` and `Jsonp` are extended by number of test to cover this new output formatting functionality. 

P.S.

Hope my improvements would be helpful! ;)


src/Http/Response.php
```diff
@@ -40,6 +40,13 @@ class Response extends IlluminateResponse
     protected static $formatters = [];
 
     /**
+     * Array of formats' options.
+     *
+     * @var array
+     */
+    protected static $formats_options = [];
```
ankitpokhrel: @vinterskogen You should use camelCase for variable names everywhere

vinterskogen: Yep, that's a relevant remark. Replaced underscore_case in variables' names to camelCase. Thanks for reviewing!

config/api.php
```diff
@@ -220,4 +220,14 @@
 
     ],
 
+    'formatsOptions' => [
```
lucasmichot: array keys should be in snake_case

src/Http/Response/Format/JsonOptionalFormatting.php
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * Determine if JSON pretty print option is set to true.
+     *
+     * @return bool
+     */
+    protected function isJsonPrettyPrintEnabled()
+    {
+        return isset($this->options['prettyPrint']) &&
+            true === $this->options['prettyPrint'];
```
lucasmichot: We don't use yoda conditions in this project

config/api.php
```diff
@@ -208,7 +208,7 @@
     |
     | Responses can be returned in multiple formats by registering different
     | response formatters. You can also customize an existing response
-    | formatter.
+    | formatter by setting options to configure it's output.
```
crhayes: All of the the instances where you changed this were originally correct.

It's = "it is"
Its = "possessive"

vinterskogen: Yeah, this is a dumb mistake. Fixed this up. Thanks for noticing.

config/api.php
vinterskogen: I think this detailed styling requirements should be added to the contributing guidelines file. :)

src/Http/Response/Format/Format.php
```diff
@@ -47,6 +54,18 @@ public function setResponse($response)
...
     /**
+     * Set the formats' options.
+     *
+     * @param array $options
+     *
+     * @return void
+     */
+    public function setOptions(array $options)
+    {
+        $this->options = $options;
```
lucasmichot: All others setters are fluent, and they return `$this`.
I suppose the same should happen here

vinterskogen: Okey. For now it is not used as fluent, but (maybe) in the future it will. I'd added this.

src/Http/Response/Format/Json.php
```diff
@@ -91,12 +98,29 @@ protected function morphToArray($value)
     /**
      * Encode the content to its JSON representation.
      *
-     * @param string $content
+     * @param array $content
      *
      * @return string
      */
-    protected function encode($content)
+    protected function encode(array $content)
```
vinterskogen: The reason I added `array` type-hinting for `$content`, is that I noticed this argument is currently marked as `string` in docbock. This is not correct, because `encode` method can be called only form number of methods in `Json` and `Jsonp` classes, and they all _explicitly_ pass array into it, and _never_ pass string. All current tests are green. Is there any possible situations, uncovered by tests, when non-array (object?...) argument can be passed into `encode` method? If the answer is "no" it not a breaking-change. (I guess it is no). Another way is to get rid of type-hinting in signature and change docblock to `@param mixed $content` or so, but I think it not a good decision, because "explicit is better than implicit". :)

config/api.php
crhayes: Sweet. English are hard 😄

src/Http/Response/Format/Format.php
```diff
@@ -47,6 +54,20 @@ public function setResponse($response)
     }
 
     /**
+     * Set the formats' options.
+     *
+     * @param array $options
+     *
+     * @return void
```
vinterskogen: Oh, right. But `@return \Dingo\Api\Http\Response\Format\Format`, as other fluent methods in this class do.

vinterskogen:
Ping @thilanga. Any pinons on this PR?

thilanga:
@vinterskogen I'll go through this today 👍 just want to check the usage . Coz there is a comment about the BC.

vinterskogen:
@thilanga Got that, thanks for replying!

thilanga:
@vinterskogen Can you update the pull request? There is a conflict

vinterskogen:
Sure. Done.

PR #1357, opened 2017-03-16
Fix Http Response parsing laravel jsonp content

Ifnot:
Hi,

**The problem** : Actually DingoApi empty all jsonp responses built with the default laravel response builder from a controller.

This works well :

```php
return response()->json([
    'foo' => 'bar',
]);
```

This return an empty response :
```php
return response()->json([
    'foo' => 'bar',
])->withCallback("test");
```

**The reason** : The `Dingo\Api\Http\Response` try to `json_decode()` the jsonp. This decode fails (because of the format of the jsonp) and return a empty json response.

**The fix** : Check if the contents contains a `/**/` at the beginning of the contents. Laravel make a `JsonResponse` for `json` and `jsonp` responses, so we cannot identify it with a class based detection. This simple check avoid a failing `json_decode`, and allows valid json to be parsed.

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -11,8 +11,8 @@ use Dingo\Api\Event\ResponseWasMorphed;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Http\Response as IlluminateResponse;
 use Illuminate\Events\Dispatcher as EventDispatcher;
-use Illuminate\Database\Eloquent\Model as EloquentModel;
 use Dingo\Api\Transformer\Factory as TransformerFactory;
+use Illuminate\Database\Eloquent\Model as EloquentModel;
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Symfony\Component\HttpKernel\Exception\NotAcceptableHttpException;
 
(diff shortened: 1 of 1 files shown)


src/Http/Response.php
```diff
@@ -95,7 +95,16 @@ public static function makeFromExisting(IlluminateResponse $old)
      */
     public static function makeFromJson(JsonResponse $json)
     {
-        $new = static::create(json_decode($json->getContent(), true), $json->getStatusCode());
+        $content = $json->getContent();
+
+        // If the contents of the JsonResponse does not starts with /**/ (typical laravel jsonp response)
+        // we assume that it is a valid json response that can be decoded, or we just use the raw jsonp
+        // contents for building the response
+        if (! starts_with($json->getContent(), '/**/')) {
```
vinterskogen: May be extract this condition to a method `doesJsonResponseContainsCallback` (or so)?

vinterskogen: May be extract this condition to method `doesJsonResponseContainsCallback` (or so)?

Ifnot: Hum,

As we are in a static method and we do not instanciate the object we should create a `protected static function`. It looks dirty.

Should i extract-it into a static anyway ? Or should we merge this solution ?

Just tell me :)

Ifnot: Any opinion on this ? :)

Ifnot:
I am pushing this again with a underlying major problem.

As dingo fail to parse the jsonp, all exceptions thrown returns a blank response if it is called from jsonp.

My PR fixes this issue, please merge-in :)

PR #1380, opened 2017-05-02
Check encode of your data

catalinux:
When a transform returns non utf8 string the response of our api is 200Ok with empty body. That is because `encode` method returns `false`.

This code will be very uses full in debugging such errors.  

Turning to PHP7 and method types would help also.

### DIFF
--- a/src/Http/Response/Format/Json.php
+++ b/src/Http/Response/Format/Json.php
@@ -2,8 +2,8 @@
 
 namespace Dingo\Api\Http\Response\Format;
 
-use Illuminate\Support\Str;
 use Illuminate\Contracts\Support\Arrayable;
+use Illuminate\Support\Str;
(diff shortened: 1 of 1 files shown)


src/Http/Response/Format/Json.php
```diff
@@ -97,6 +97,11 @@ protected function morphToArray($value)
      */
     protected function encode($content)
     {
-        return json_encode($content);
+        $encodedString = json_encode($content);
+        if ($encodedString === false) {
+            throw new \ErrorException('Error encoding data in JSON format: '.json_last_error());
```
lucasmichot: You should import `\ErrorException` and add a `@throws \ErrorException` docblock to this method

lucasmichot:
Tests ?

PR #1384, opened 2017-05-07
Fix 'Illegal offset type' error for non-object argument passed to hasBinding

meysampg:
In situations that the parameter passed to `hasBinding` of Transformer for example is an array (suppose paginate a set of arrays and then the first() of the collection is an array and it's illegal for offset of ::$bindings) and it causes a fatal error. This pull request fix it.

### DIFF
--- a/src/Transformer/Factory.php
+++ b/src/Transformer/Factory.php
@@ -176,7 +176,7 @@ class Factory
 
         $class = is_object($class) ? get_class($class) : $class;
 
-        return isset($this->bindings[$class]);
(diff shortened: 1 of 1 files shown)


thilanga:
@meysampg Can I have tests for this?

thilanga:
@meysampg Any update on this?

meysampg:
As I rechecked documentation, the problem which this pull request tries to solve it, can be solved by creating an explicit transformer and in fact that's not a bug for your library. So I close this pull request.

PR #1394, opened 2017-05-25
Add possibility to return json data as objects rather than arrays

rundef:


rundef:
I encountered issues with this, so I'm calling a function to cast the response to an object instead.

```php
function arrayToObject($arr) {
    return json_decode(json_encode($arr), false);
}
```

PR #1408, opened 2017-07-10
Allow integration with Lumen-Route-Binding

mmghv:
Set dispatcher for Lumen by resolver callback.

This will allow us to easily pass a custom dispatcher for Lumen by extending the service provider, needed to allow integration with [lumen-route-binding](https://github.com/mmghv/lumen-route-binding), A package to add support for `Route Model Binding` in Lumen.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: php
 
 sudo: false
+dist: trusty
 
 env:
   global:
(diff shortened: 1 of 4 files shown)


bradders:
Any ideas on when (or if) this will make it in? I'm currently adding Dingo to an existing lumen project that uses route binding to type hint my routes. It would be great if they worked together nicely.

mmghv:
@bradders until this is accepted, you can follow these steps to make it work.

bradders:
@mmghv So good, thanks for that!

PR #1412, opened 2017-07-26
fix: Call the correct session setter in Laravel 5.4

eddiclin:
The http session is refactored in Laravel 5.4. The session handlers no longer implements Symfony's SessionInterface. 
Check links as below,
https://github.com/laravel/framework/pull/17064
https://laravel.com/docs/5.4/upgrade

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: php
 
 sudo: false
+dist: trusty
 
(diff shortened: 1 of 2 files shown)


dmason30:
@thilanga Can we not get a 5.4 compatible release with this fix in it?

PR #1451, opened 2017-09-20
Laravel 5.5 support

thilanga:
- Laravel 5.5 support
- Removed broken OAuth2 provider with tests
- New github issue template
- StyleCI fixes
- Dingo Router extends form the Illuminate route


from2day:
Check. All good. Merge and I'll test it.

crynobone:
There's no guarantee there no breaking change between Route for Laravel 5.1, 5.2, 5.3, 5.4 and 5.5.

You sure want to extends and at the same time supports such varies version?

thilanga:
First thing first we need to support Laravel 5.5. after this release. I'll work on below 5.5. at the moment it's too much. If someone wants to use the package against 5.4 and below beta8 is there

vrkansagara:
yes @thilanga  👍

Issue #1275, opened 2016-12-21
Middleware terminate() not being called


diegofalciola:
Reopening issue #721

MBoretto:
Any update on this? I have this issue on:
```
laravel/lumen-framework           v5.3.3  The Laravel Lumen Framework.
```

k-szulka:
me to, when is this going to be fixed

matrixn:
same issue with lumen 5.4 ...any news on this?

PR #1485, opened 2017-11-25
fix:Middleware terminate() not being called

yangzuwei:
I try to fix the issue:https://github.com/dingo/api/issues/1275


yangzuwei:
\vendor\dingo\api\src\Http\Middleware\Request.php:line 125
remove ->through($this->middleware),avoid execution twice

PR #1494, opened 2017-12-07
I think we can remove the method "through()" from here

yangzuwei:
Because the "handle" method also be executed in Lumen


tripper54:
This change means middleware defined in the $middleware of Kernel.php does not run. Why make this change?

LarryZhang:
this pull should disable other middlewares in laravel

Issue #1496, opened 2017-12-11
FormRequest Validation - does not describe which fields failed validation


simplenotezy:
| Q                 | A
| ----------------- | ---
| Bug?              | yes
| New Feature?      | no
| Framework         | Larave
| Framework version | 5.5.*
| Package version   | 2.0.0-alpha1
| PHP version       | 7.1.10

#### Actual Behaviour
When using FormRequest to validate input given by client, and validation fails, it simply returns:

    "message": "The given data was invalid.",

Along with status_code 500 + debug trace.

#### Expected Behaviour

I would expect it to give some more clues as to which validations that failed, and why.

simplenotezy:
Solution, in your `FormRequest` use

    use Dingo\Api\Http\FormRequest;

Instead of

    use Illuminate\Foundation\Http\FormRequest;

danijelk:
Sitting in the same boat, tried to do abstract class Form extends \Dingo\Api\Http\FormRequest, what error should be thrown?

Tried: $this->throwValidationException(request(), $validator) and: $this->failedValidation($validator)

yangliulnn:
But the status code is still 500

danijelk:
Indeed this should most definitely not be a 500

yangliulnn:
#1519

specialtactics:
Closing, PR was merged.

PR #1510, opened 2018-01-13
Show errors for ValidationException

jianfengye:
I got the same issue like https://github.com/dingo/api/issues/1496

And I check Laravel source Code 

ValidateException has errors() method, it will show which field has validete error. And dingo/api‘s :error does not got this errors, that will make user unreachable the true validate error.

so I add this pr for show errors for ValidationException. and the response will like that:
image


keyganker:
@thilanga @jianfengye This PR cause error under laravel `5.1, 5.2, 5.3, 5.4`, because the `ValidationException->errors` method doesn't exists before laravel 5.5

PR #1511, opened 2018-01-15
Change paginator to output json format. 

qufo:
We use dingo/api to build api site , when use paginator function, the response output did not include previous page  url and next page url . so add toJson() to $paginator will fix this.


qufo:
Sorry, All my fault. I'll close it.

PR #1532, opened 2018-02-25
Change HTTP Status Code on Validation ex

SamsamBabadi:
ValidationException has status code for response but does not implement HttpExceptionInterface.
Status code by default is 422 but can change it


thilanga:
@SamsamBabadi Can you update the PR with UnitTests?

SamsamBabadi:
I will working on UnitTests as soon as possible.

PR #1534, opened 2018-03-01
Fix FormRequests missing class error

gazben:
Lumen does not contain the Illuminate\Foundation\Http\FormRequest class.

I merged the original file and the Laravel FormRequest class into the Dingo FormRequest class.

I added the ValidatesWhenResolvedTrait resolving to the LumenServiceProvider too.

### DIFF
--- a/src/Http/FormRequest.php
+++ b/src/Http/FormRequest.php
@@ -143,7 +143,6 @@ class FormRequest extends Request implements ValidatesWhenResolved
      * @param  \Illuminate\Contracts\Validation\Validator  $validator
      *
      * @return void
-     *
      */
     protected function failedValidation(Validator $validator)
     {
@@ -158,6 +157,7 @@ class FormRequest extends Request implements ValidatesWhenResolved
      * Get the proper failed validation response for the request.
      *
      * @param  array  $errors
+     *
      * @return \Symfony\Component\HttpFoundation\Response
      */
     public function response(array $errors)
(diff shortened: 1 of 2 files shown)


sakalauskas:
@thilanga @gazben FYI this PR breaks dingo on Laravel. 

In LaravelServiceProvider we need to add ```$this->app->resolving(FormRequest...``` too, but problem is that we cannot set the redirector property as in Laravel class is named "Illuminate\Routing\Redirector" and in Lumen it is named "Laravel\Lumen\Http\Redirector" and unfortunately there is no Redirector contract. Thus, I am not sure at the moment how to resolve this nicely.

gazben:
@sakalauskas In my experience it "only" breaks in the latest laravel release. 
I will look into this today/tomorrow.

TeroBlaZe:
@gazben Can't use an application after this PR. Validation fails on POST request with json. Rolled back to previous commit.

TeroBlaZe:
@gazben Laravel

gazben:
@TeroBlaZe @sakalauskas I can only present an ugly solution:
https://github.com/Brixify/api/commit/e6a2d2cd5fa3e42954e6358d364375dbd38adea2

Without a Redirector contract I don't think I can do more. What do you think? (works on my machine)

gazben:
@TeroBlaZe @sakalauskas I created another pull request for the issue.

sakalauskas:
@gazben This looks fine to me.

PR #1537, opened 2018-03-08
Add getMiddleware function to Route class

gazben:
The getMiddleware function is missing and reported here: https://github.com/flipboxstudio/lumen-generator/issues/25
Also I think you have to be able to query the middlewares.

If the function is not necessary and you can query the middlewares with another method, please write it in the comments.

### DIFF
--- a/src/Routing/Route.php
+++ b/src/Routing/Route.php
@@ -533,4 +533,14 @@ class Route extends \Illuminate\Routing\Route
     {
         return in_array('https', $this->action, true);
     }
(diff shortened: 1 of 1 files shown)


telefoontoestel:
I noticed that this issue has not been released yet and I would like to submit a suggestion. Within the Laravel own Route class there is a method called gatherMiddleware that collects the middleware of the route. Shouldn't the method at least be called the same, since it is used in the same manner?

PR #1545, opened 2018-03-27
Route cache service injection

bjdelange:
The Routes command has pretty constructor service injection; so does the Cache command now.
Also cleaned up a double used variable.


specialtactics:
Hi @bjdelange - looks good to me, can you please sync your master with the repo which will fix the styleCI check

bjdelange:
I didn't have my copy anymore so I re-created it here: #1618

PR #1549, opened 2018-04-17
fix ValidationException->errors() method not exists for laravel 5.2,5.3,5.4

keyganker:


specialtactics:
Based on some earlier commits by @thilanga , this package doesn't support laravel below 5.5 in general.

It's possible possible that this will fix this particular issue, but if we continue to support laravel version this old, it will cause a lot of messy code I think.

It would be much better if users updated their versions of laravel, they are generally very easy to update, and there are even automated services that can do this 

One example: https://laravelshift.com/

keyganker:
Thanks for your relay!

PR #1550, opened 2018-04-18
not is api prefix , no dingo api loading

fdwl:


specialtactics:
Hi @fdwl 

Can you explain what's going on there, and the use case?

Schnoop:
I think he tries to fix something i mentioned here: https://github.com/dingo/api/issues/1647

specialtactics:
It is definitely not necessary to have an API prefix for dingo to work, so this change will break a lot of setups, and I don't really understand why you would want it in the first place.

Issue #1552, opened 2018-04-19
Framework new version incompatibility


dogthief:
| Q                 | A
| ----------------- | ---
| Bug?               |no
| New Feature?     | yes
| Framework         | Lumen
| Framework version | 5.6.3
| Package version   | 2.0.0
| PHP version       | 7.x.y

#### Actual Behaviour

Skeleton revision is currently incompatible
#### Expected Behaviour                                                                                                                                                                                                                               Type error: Argument 1 passed to Illuminate\Routing\RouteCollection::add() must be an instance of Illuminate\Routing\Route, null given, called in /var/www/service.wow-trend.com/vendor/illuminate/routing/ResourceRegistrar.php on line 99  
                                                                                                                                                                                                                                            
#### Steps to Reproduce
As long as the binding resource route is abnormal
api->resource('/abilities', 'AbilitiesController');


#### Possible Solutions

specialtactics:
Something which the documentation probably does not make very clear;

When you put this façade in your app:

```php
Dingo\Api\Facade\Route 
```

Do not give it the alias "Route" - ie. the same as the laravel/lumen router. Call it something else (like APIRouter).

Git-yyc:
Is an updated version of the Illuminate \ Routing library. The new version of the RouteCollection::add method adds a return value, causing the current version of dingo/api to be incompatible

Git-yyc:
@specialtactics

specialtactics:
I see, no problem I have fixed that and made a PR #1555

ldanielduarte:
Laravel version: 5.6.20
dingo version : 1.0.0-beta8;
php version: 7.1.6

While migrating Laravel version to 5.6, I'm having a problem with the api urls with aliases generation:

`Error : Call to undefined method Dingo\Api\Routing\Route::getDomain()`

I think this is a problem of incompatibility with the new https://github.com/laravel/framework/blob/5.6/src/Illuminate/Routing/RouteUrlGenerator.php 

Is anyone else having this problem?

PR #1555, opened 2018-05-02
Laravel compatibility with RouteCollection::add()

specialtactics:
Return Route from RouteCollection::add() for upstream compatibility with Laravel Router 
#1552

### DIFF
--- a/src/Routing/RouteCollection.php
+++ b/src/Routing/RouteCollection.php
@@ -34,13 +34,15 @@ class RouteCollection implements Countable, IteratorAggregate
      *
      * @param \Dingo\Api\Routing\Route $route
      *
-     * @return void
+     * @return \Dingo\Api\Routing\Route
      */
     public function add(Route $route)
     {
         $this->routes[] = $route;
 
         $this->addLookups($route);
+
+        return $route;
     }
 
     /**


catalinux:
is there a reason for not merging this pull request?

catalinux:
ping @thilanga

thilanga:
I'll go through the PR today and get it merge

catalinux:
Great news. Thanks for all the effort!

thilanga:
@catalinux @specialtactics Can I get this green before the merge please?

specialtactics:
@thilanga Hey there, I had a look and it seems like there is a config issue with StyleCI - not sure I am able to fix that as it would require owners to fix it?

I will have a look at Travis, I assumed it was the same error but now see there is actually a failing test.

catalinux:
@specialtactics I think it fails a test
https://travis-ci.org/dingo/api/jobs/373878732

```
1) Dingo\Api\Tests\Routing\Adapter\LumenTest::testRoutingResources
TypeError: Argument 1 passed to Illuminate\Routing\RouteCollection::add() must be an instance of Illuminate\Routing\Route, null given, called in /home/travis/build/dingo/api/vendor/illuminate/routing/ResourceRegistrar.php on line 99
/home/travis/build/dingo/api/vendor/illuminate/routing/RouteCollection.php:50
/home/travis/build/dingo/api/vendor/illuminate/routing/ResourceRegistrar.php:99
/home/travis/build/dingo/api/src/Routing/Router.php:320
/home/travis/build/dingo/api/src/Routing/Router.php:299
/home/travis/build/dingo/api/tests/Routing/Adapter/BaseAdapterTest.php:195
/home/travis/build/dingo/api/src/Routing/Router.php:171
/home/travis/build/dingo/api/src/Routing/Router.php:136
/home/travis/build/dingo/api/tests/Routing/Adapter/BaseAdapterTest.php:197

specialtactics:
@catalinux @thilanga Yep, I just had a look, it seems like the test actually fails regardless of this change or not - ie. If you just check out master of dingo/api and run the tests, this particular test will fail. So I am guessing there was an update which broke it in Laravel/Lumen?

I just a bit of a play around with it, and it seems like the basic problem is that there is a test for Lumen (the test failing), whereby a route is being added using the lumen adapter, but laravel code that is running is expecting a route object to be returned, where one won't be (because the lumen function for addRoute differs from the Laravel one in terms of not returning anything, whereas the Laravel function returns the route object).

I am not entirely sure how this situation came about - perhaps in the past Laravel also did not return the route, so nothing was expected there?

I think the test shouldn't be running through Laravel to the extent that it is, because in the call tree you can see Laravel and Lumen vendor functions being mixed, and this is the cause of the problem (ie. they are not compatible when it comes to routing at least).

Not sure what the next step is, I can try and work with it more, but would like to know a bit about the test setup and why it is doing what it's doing.

specialtactics:
Actually after looking at it a bit further, I would suggest this package is just not compatible with Lumen anymore.

gazben:
@specialtactics If you add the old method to vendor/dingo/api/src/Routing/ResourceRegistrar.php it works. 
I will investigate, why this change was made, and where does the new return value go.

gazben:
@thilanga After a bit of investigation I did not find any usage of the returned value in the laravel repo so I think the merge of #1580 is safe. The linting and unit tests are also fixed there, so if you first merge #1580 and after #1555 all checks will be green

thilanga:
Beautiful. I'll do this as soon as I go to work.💪

specialtactics:
Thanks guys! Sorry for slow follow-up, I was away

PR #1564, opened 2018-06-04
Update Factory.php

pimplesushant:
Added method to throw unprocessableEntity exception with status code 422.

### DIFF
--- a/src/Http/Response/Factory.php
+++ b/src/Http/Response/Factory.php
@@ -267,7 +267,7 @@ class Factory
      *
      * @return void
      */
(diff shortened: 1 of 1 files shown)


gazben:
@pimplesushant Can you rerun the CI please? The configuration is fixed now

specialtactics:
To keep consistency with the rest of the error responses, probably the function should be called "errorUnprocessableEntity"

specialtactics:
Awesome @pimplesushant all looks well. Only other problem I just noticed is that the styleCI check is failing - but that is a simple fix, just sync your girhub repo with master, and it should pass.

Issue #1133, opened 2016-07-25
Transformer optional include must be a model attribute


jarednipper:
I have an optional include on a transformer like so:

```
public function includeContributingUsers(Conversation $conversation)
{
    $contributors = $this->conversationService->getContributors($conversation);
    return $this->collection($contributors, new UserTransformer);
}
```

This works fine if I am transforming a single conversation as an Item. But when I try to transform a Collection of conversations I get: 

`Call to undefined method Illuminate\Database\Query\Builder::contributingUsers()`

It is looking for the relationship on the model, but it is supposed to be calling the `includeContributingUsers()` method on the transformer.

This Fractal issue describes the same problem: https://github.com/thephpleague/fractal/issues/240

Another: https://github.com/dingo/api/issues/1096#issuecomment-230889539

georaldc:
I believe this happens because of implicit eager loading by Dingo when using "includes" and when a collection is encountered:

https://github.com/dingo/api/blob/master/src/Transformer/Adapter/Fractal.php#L89

Not sure of a way around this though, other than defining the method on your model or probably disabling automatic eager loading and handling it yourself when needed.

tr33m4n:
I had this issue a while back as well, I had to add the method to the model and call a fake relationship

hskrasek:
@jarednipper @georaldc and @tr33m4n are correct. You'd have to call the fractal adapter, and disable the eager loading feature.

PDeveloper:
@tr33m4n Same here. Added a duplicate / fake relationship. Striving for cleaner code, this is not ideal.

JonasDoebertin:
Would be nice if we were able to specify which includes should be eager-loaded right within each transformer to be able to have includes that don't resolve to a relationship directly.

Maybe as `protected $eagerLoadedIncludes = ['user', …];`

piyushkantm:
Any updates on this?

vruello:
If anyone is still interested in resolving this issue, I have made a quick fix using the idea of @JonasDoebertin .

I've added `public $lazyLoadedIncludes = ['foo'];` into my transformer, and changed the `mergeEagerLoads` method of `Dingo\Api\Transformer\Abstract\Fractal` into : 

```
 /**
     * Get includes as their array keys for eager loading.
     *
     * @param \League\Fractal\TransformerAbstract $transformer
     * @param string|array                        $requestedIncludes
     *
     * @return array
     */
    protected function mergeEagerLoads($transformer, $requestedIncludes)
[...]

PR #1575, opened 2018-07-15
Support for custom data includes in transformer

vruello:
Quick fix for #1133 and related issues. 

The attribute `lazyLoadedIncludes` must be declared **public** in the transformer to make it work. Otherwise, it would have been necessary to change Fractal library to add a getter.

All the includes name listed in this array will not be loaded using eager loading. It is then possible to include custom data not linked to any relationship or model.


gazben:
@vruello I resubmitted your changes now the CI will accept it.

PR #1581, opened 2018-08-16
Fix Style CI

gazben:


gazben:
@thilanga After this merge I think you should create a new release, because the current one should be broken because of the upstream changes

thilanga:
@gazben I'll have a look.

PR #1590, opened 2018-09-07
Fixed issue on URLGenerator for signed routes

harlekoy:


specialtactics:
@harlekoy I just made a slight update to comply with how Laravel core is doing this (for better backwards compatibility), and merged !

PR #1595, opened 2018-09-26
Fixes a problem whereby parameters passed through internal requests a…

specialtactics:
This fixes the following problem -

If you create an internal request, and you pass it parameters, these parameters end up in the $request->request ParameterBag. The problem is, Laravel's $request->input() will check Laravel's $request->json ParameterBag instead, and thus these parameters will not be accessible using Laravel's $request->input() method.

As a workaround it is possible to access them by using $request->request->get(), however Laravel recommends not using that and instead using it's own $request->input() function. Consequently existing middleware written by and for laravel is built to use $request->input(), and most Laravel code written uses $request->input().

For consistency, so as to not introduce problems with conflicting parameter getting from a request, this small change addresses this problem by passing through request parameters into Laravel's json ParameterBag


specialtactics:
Not sure what's happening with styleCI here. I removed the linting line because it was coming up with an error that it's invalid, and according to their docs it was removed due to being deprecated for some time. Now it seems to time out or something.

specialtactics:
@thilanga Fixed styleCI, can you please check this out?

specialtactics:
@jasonlewis Just pinging you as well to have a look m8

pawlox:
Hey, what is a status of this pull request? Can someone merge it?

specialtactics:
Merged !

PR #1610, opened 2018-12-28
Update .gitattributes

lloricode:


specialtactics:
Why is .styleci.yml in there?

lloricode:
because it doesn't need on production

specialtactics:
@lloricode This is a repository for development, it has nothing to do with your own production environment.

lloricode:
@specialtactics i think your right :)

PR #1611, opened 2018-12-29
Fixed an issue with no content

ChienDevIT:
Fixed "sha1() expects parameter 1 to be string, null given" when returning empty content


specialtactics:
Thanks for your contribution @ChienDevIT, merged !

Issue #1608, opened 2018-12-12
Controller Constructor called tw


crxgames:
| Q                 | A
| ----------------- | ---
| Bug?              |yes
| New Feature?      | no
| Framework         | Laravel
| Framework version | 5.3+
| Package version   | 2.0.0-alpha1
| PHP version       | 7.2.11

#### Actual Behaviour

My controller constructors are being called twice. It appears that there are two different instantiations of the Controller taking place upon execution.

#### Expected Behaviour

One instance of controller being instantiated.

#### Steps to Reproduce

1. Laravel 5.7 installation + dingo 2.0.0-alpha1
2. Create new controller and define a constructor with an echo statement
3. Add route to new controller method.
4. Load page at route defined above. You will see constructor output twice.

#### Possible Solutions

This is probably not correct, but currently I have had to convert Route.php line 283 to:
        $this->container->instance($this->controllerClass, $this->controller);

specialtactics:
This is an interesting one.

I did a bunch of testing and investigation, and essentially Laravel and Dingo both go off dong their own routing and controller instantiation - difference being, only the dingo controller actually ends up being used.

Yet if Laravel did not do it's own thing, there would be other problems with the request, so probably that is fine. 

Because Laravel instantiates the controller first, and I see no difference between them, I added a resolved check in the dingo router to not instantiate a controller if it has been already, and everything _seems_ to still work fine.

I will keep that PR open for a while, just to get some feedback and merge if nobody picks up any issues.

I recommend you give it a try as well.

https://github.com/dingo/api/pull/1614

specialtactics:
Just an updated, seems like that solution won't work (no change to current situation), because Laravel doesn't bind the controller instance when it instantiates it.

I will think of other possible solutions for this, but for now potentially you could keep a static variable in your controller class, to denote whether it's been instantiated or not, if there is anything in your controller constructor you do not wish to run more than once.

PR #1614, opened 2019-01-25
Instantiate new controller only if not resolved already

specialtactics:
This deals with issue https://github.com/dingo/api/issues/1608

I am not sure as to whether something has changed in Laravel which led to this problem, or if it's always been the case.

This change passed a large number of unit tests I have set up for projects using dingo, so I don't see any problem with this change, but looking for other people's feedback on whether they can sense any issues - especially on older laravel vesions.


specialtactics:
Seems like the reason it isn't working is that laravel doesn't bind the controller when it instantiates it. I am guessing there are good reasons for this, and not something we can change anyway.

I will have a think about other possible solutions.

PR #1618, opened 2019-02-11
 Route cache service injection

bjdelange:
The Routes command has pretty constructor service injection; so does the Cache command now.
Also cleaned up a double used variable.

### DIFF
--- a/src/Console/Command/Cache.php
+++ b/src/Console/Command/Cache.php
@@ -2,9 +2,10 @@
 
 namespace Dingo\Api\Console\Command;
 
+use Dingo\Api\Routing\Router;
 use Illuminate\Console\Command;
 use Illuminate\Filesystem\Filesystem;
-use Illuminate\Contracts\Console\Kernel;
+use Dingo\Api\Contract\Routing\Adapter;
 
 class Cache extends Command
 {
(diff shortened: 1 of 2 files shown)


specialtactics:
Hey @bjdelange I appreciate you taking the time to re-submit this, thanks for your work !

src/Console/Command/Cache.php
```diff
@@ -52,15 +71,13 @@ public function handle()
     {
         $this->callSilent('route:clear');
 
-        $app = $this->getFreshApplication();
```
crynobone: Why would you remove this. 

```
$this->callSilent('route:clear'); 
```

This would clear the cache route, if you don't get a fresh application wouldn't `$app` already contained route from cached? This why even Laravel itself get a refresh application instance on route caching command.

https://github.com/laravel/framework/blob/5.7/src/Illuminate/Foundation/Console/RouteCacheCommand.php#L55

specialtactics: This is at least my understanding but, it's a bit redundant precisely because Laravel already does this and caches the up-to-date routes, which is done by the 

```php
        $this->callSilent('route:clear');
```

Keen to hear from author though.

bjdelange: I made this PR a long time ago and don't remember the details that accurately anymore.
Shortly afterwards we removed Dingo from our setup altogether (too little benefit for the performance cost).
In the meantime I've changed jobs and don't work with Laravel at all anymore.
Sorry I can't be of any more help anymore.

specialtactics: Okey, I might just add that back just in case for now.

Issue #1232, opened 2016-10-17
Route Patterns Not Working


bbrink68:
Laravel 5.3
Dingo   1.0.*@dev

I've read in these issues: https://github.com/dingo/api/issues/741 & https://github.com/dingo/api/issues/724

That global defined patterns should work with Dingo API Router, but this is not the case.

``` php
    class RouteServiceProvider extends ServiceProvider {

        public function boot()
        {
            // Placed here - not working
            // Route::pattern('id', '[\d+]');

            parent::boot();
        }

        public function map()
        {
            // Tried here too, doesn't work.
            // Route::pattern('id', '[\d+]');

            $this->mapApiRoutes();

            $this->mapWebRoutes();
        }
    }
```

Has anyone gotten patterns to work with Dingo? How did you do it since I've read that it should work.

I've also tried injecting `Router` into the `map()` and `boot()` methods, still a no go.

bbrink68:
Just found out that I needed to inject Laravel Router into the `map()` method in `RouteServiceProvider` like:

```
public function map(Router $router)
{
    $router->pattern('id', '[\d+]); // <-- Works
    Route::pattern('id', '[\d+]');    // <-- Does Not Work
}
```

alexjose:
+1

jasonlewis:
Can confirm they aren't working.

From a quick glance looks like the router patterns need to be merged in the Laravel routing adapter when the route is created.

bitdagger:
Ran into this today myself. I was wondering why my route patterns weren't working, then I realized those routes were Dingo, not Laravel's router. A quick search later led me here.  

+1

mehranhadidi:
so any fix? any suggestion?

ntopulos:
Did anyone find some solution to this problem?

specialtactics:
Hey guys

I've implemented this in the following PR, can you please try it and see if it works for you, without any unexpected results?

https://github.com/dingo/api/pull/1619

christoph-kluge:
Hey @specialtactics, thanks for the support. I had this issue in the past and ignore it for a long time. Now I've tackled it again and discovered that we need to add the patterns already when adding the route to the adapter. Otherwise it will take the first route which is wrong. Here is an example:

```
$api->get('/entity/{uuid}')
$api->get('/entity/{slug}')
$api->get('/entity/{customIdentifier}')
```

I have a local working draft and forked the repo already. Will add the required changes soon.

specialtactics:
[...]

PR #1619, opened 2019-02-12
Feature/implement param pattern matching

specialtactics:
For issue #1232

I spent a bit of time today looking at this, and it seems like there are very big deviations between Laravel's and Dingo's routing workflow. Whether this is by design or not... I really don't know. Anyway, I implemented this in a fairly simple way after some digging and digging, and it does seem to work for me, and passed a lot of unit tests in some large projects.

At the same time, the functionality is not 100% compatible with Laravel's Router - which checks each route individually, including this param pattern matching (until it finds the matching route), whereas this implementation does the same without param pattern matching, and then does that validation for the one found route at the end.

This could lead to inconsistent results if there are very ambiguous and routes defines (which is rather sloppy in any case).

Keen to have other people give this a try and give me feedback before I merge.

### DIFF
--- a/src/Routing/Adapter/Laravel.php
+++ b/src/Routing/Adapter/Laravel.php
@@ -47,6 +47,13 @@ class Laravel implements Adapter
     protected $oldRoutes;
 
     /**
+     * The globally available parameter patterns.
+     *
+     * @var array
+     */
+    protected $patterns = [];
+
+    /**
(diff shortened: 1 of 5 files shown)


specialtactics:
I've done fair bit of testing and this seems to work, merging now due to lack of feedback.

PR #1623, opened 2019-02-26
Update Router.php, var $version correct data type is array|string

maxsky:
fix `$version` data type of the `version()` support array.


specialtactics:
Thanks @maxsky

PR #1625, opened 2019-02-28
Add new methods to satisfy Application contract for Laravel 5.8

### LINKED ISSUE #1624: namespace Dingo\Api\Exception not work in laravel 5.8
laravel-framework file 
`namespace Illuminate\Contracts\Debug `
has upgrade ,
this interface add this function 
`    public function shouldReport(Exception $e);`
but This interface is not implemented
please check and fix it. 
thx.

mallardduck:
## PR Not Final
Until Lumen publishes a tag for 5.8 the testing framework used for this package is broken. So the main option is to wait until that happens, with an alternaitive of simply modifying the dependency for lumen-framework to not use `dev-master`.

## Intent of PR

This PR adds all the new methods for Laravel 5.8 as required by the updated contract.

Fixes #1624

### Notes

This change also includes some necessary modifications to how testing works, so as to allow the tests to support Laravel 5.5-5.7 and 5.8. The package itself seems unaffected, just the tests were having issues.

### DIFF
--- a/src/Exception/Handler.php
+++ b/src/Exception/Handler.php
@@ -82,13 +82,14 @@ class Handler implements ExceptionHandler, IlluminateExceptionHandler
      * Determine if the exception should be reported.
      *
      * @param  \Exception  $e
+     *
      * @return bool
      */
     public function shouldReport(Exception $e)
     {
         return true;
     }
-    
+
     /**
      * Render an exception into an HTTP response.
      *
--- a/src/Exception/Handler.php
+++ b/src/Exception/Handler.php
@@ -79,6 +79,17 @@ class Handler implements ExceptionHandler, IlluminateExceptionHandler
     }
 
     /**
+     * Determine if the exception should be reported.
+     *
+     * @param  \Exception  $e
+     * @return bool
+     */
+    public function shouldReport(Exception $e)
+    {
+        return true;
+    }
+    
+    /**
      * Render an exception into an HTTP response.
      *
      * @param \Dingo\Api\Http\Request $request
--- a/tests/ChecksLaravelVersionTrait.php
+++ b/tests/ChecksLaravelVersionTrait.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace Dingo\Api\Tests;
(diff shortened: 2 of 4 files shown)


mallardduck:
This only fails testing because of API changes in Laravel 5.7 vs 5.8.

I would make an attempt to resolve the errors, however there are too many ways this could be addressed for me to attempt it. I need some direction from maintainers on how this should be addressed.

The common option is to tag a specific version of the package for the new laravel version. So V2 of dingo is 5.5-5.7 and V3 of Dingo would be 5.8+ (at least until another breaking change).

Or since it's literally just a test stub file causing it to fail, we could adjust how tests are configured to detect which stub to use. Then tests could just switch between the pre-5.8 and post-5.8 stubs.

mallardduck:
Update: This PR will not fix all the issues - it merely fixes the first error, but unfolds many many more. The Application contract was updated in Laravel 5.8, even if we make changes to address these changes it won't fix the issue.

Details here: https://laravel.com/docs/5.8/upgrade#the-application-contract

Even after making the necessary changes to the Stub file, we'll run into a new error due to Lumen not being fully updated for 5.8 yet. Essentially changes in `illuminate/support` have been made and Lumen hasn't been updated to work with them yet. So the `dev-master` branch of Lumen is partially there, but not quite.


The [offending line](https://github.com/laravel/framework/blob/4f0a89f9d6f02c108cc400a2f1f0eeaea3e25a87/src/Illuminate/Support/helpers.php#L645).

specialtactics:
@mallardduck Thanks for your contributions here.

In terms of the application stub, why not just update the existing one? If I try you 58 stub as replacing the normal one, all the tests still pass?

So far as Lumen, I actually think we should just drop it's support going forward, as it doesn't seem to be that it is being used frequently for new projects within the community nor supported well by the Laravel team themselves.

specialtactics:
Excluding the Lumen routing tests has corrected the testing not working. I will merge this PR for now because it's great, but let me know your thoughts on the two application stubs vs one.

PR #1637, opened 2019-03-19
Use Cache TTLs using now() - Laravel 5.8 compatibility

specialtactics:
Bump up minimum required version of Laravel/Lumen to 5.5

Pull in Laravel/framework. The reason for this is to access the now() helper. Unfortunately because Laravel 5.8 introduces some quite flexible date handling and the ability to use date handlers even different to Carbon, makes it complex to port the now() function helper in isolation, while maintaining compatability with laravel 5.8+, so it is a better option to require the whole framework.

In practise, this should only affect users of Lumen, to the extent that this will load the laravel framework helpers.


specialtactics:
Going to tag this as 2.2.0

PR #1640, opened 2019-03-20
Missing use of Kernel

timjuly:
missing use of kernel in commit 398b4612a6befc24efeab3a629dbed12c0531704


timjuly:
@specialtactics

specialtactics:
Thanks @timjuly ! Tagged

PR #1644, opened 2019-04-04
Improving Response Factory Doc

stephandesouza:
`$tranformer` accepts `string|callable|object` instead only of `object`


specialtactics:
Thanks @stephandesouza

PR #1654, opened 2019-05-06
FormRequest->withValidator is not working as validateResolved is not getting called.

### LINKED ISSUE #1653: FormRequest->withValidator is never getting called
| Q                 | A
| ----------------- | ---
| Bug?              | yes
| New Feature?      | no
| Framework         | Lumen
| Framework version | 5.8.0
| Package version   | 2.2.3
| PHP version       | 7.2.14

#### Actual Behaviour

`withValidator` function gives us the ability the customise the validator instance before evaluating the rules. But [withValidator](https://github.com/dingo/api/blob/8f97fadae800202e618e0c3f678b4aab78e1dc05/src/Http/FormRequest.php#L105) is never getting called, because in [LumenServiceProvider](https://github.com/dingo/api/blob/14225351b2a9b652f6814851539370645c0c1fac/src/Provider/LumenServiceProvider.php#L58), we are directly calling the [$resolved->validate()](https://github.com/dingo/api/blob/8f97fadae800202e618e0c3f678b4aab78e1dc05/src/Http/FormRequest.php#L75) which never calls the `FormRequest->getValidatorInstance` method.

#### Expected Behaviour

`withValidator` should be called before evaluating the rules.

#### Steps to Reproduce

I am using the `withValidator` function to validate the `current_password` field when a `Change Password` request comes.
```php
<?php
namespace App\Http\Requests;

use Dingo\Api\Http\FormRequest;
use Illuminate\Validation\Validator;

class ChangePasswordRequest extends FormRequest
{
  public function authorize(): bool {
    return true;
  }

  public function rules(): array
  {
    return ["current" => "required|string", "password" => "required|string|confirmed"];
  }

  public function withValidator (Validator $validator): void
  {
    $validator->after(function ($validator) {
      if (!app('hash')->check($this->current, $this->user()->password)) {
        $validator->errors()->add("current", "Current password is incorrect");
      }
    });
  }
}
```
Here, `withValidator` function is never getting called and validation passes even when the current password in not correct.

#### Possible Solutions

As described in [FormRequestServiceProvider](https://github.com/laravel/framework/blob/e6c8aa0e39d8f91068ad1c299546536e9f25ef63/src/Illuminate/Foundation/Providers/FormRequestServiceProvider.php#L30), we should call `$resolved->validateResolved()` in the LumenServiceProvider.

sudkumar:
Fixes #1653

### DIFF
--- a/src/Console/Command/Routes.php
+++ b/src/Console/Command/Routes.php
@@ -144,7 +144,7 @@ class Routes extends RouteListCommand
      */
     protected function routeRateLimit($route)
     {
-        list($limit, $expires) = [$route->getRateLimit(), $route->getRateLimitExpiration()];
+        [$limit, $expires] = [$route->getRateLimit(), $route->getRateLimitExpiration()];
 
         if ($limit && $expires) {
             return sprintf('%s req/s', round($limit / ($expires * 60), 2));
--- a/src/Http/Middleware/Request.php
+++ b/src/Http/Middleware/Request.php
@@ -157,7 +157,7 @@ class Request
                 continue;
             }
 
-            list($name, $parameters) = $this->parseMiddleware($middleware);
+            [$name, $parameters] = $this->parseMiddleware($middleware);
 
             $instance = $this->app->make($name);
 
@@ -178,7 +178,7 @@ class Request
      */
     protected function parseMiddleware($middleware)
     {
-        list($name, $parameters) = array_pad(explode(':', $middleware, 2), 2, []);
+        [$name, $parameters] = array_pad(explode(':', $middleware, 2), 2, []);
 
         if (is_string($parameters)) {
             $parameters = explode(',', $parameters);
--- a/src/Provider/LumenServiceProvider.php
+++ b/src/Provider/LumenServiceProvider.php
@@ -55,7 +55,7 @@ class LumenServiceProvider extends DingoServiceProvider
         });
 
         $this->app->afterResolving(ValidatesWhenResolved::class, function ($resolved) {
-            $resolved->validate();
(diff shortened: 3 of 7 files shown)


sudkumar:
@specialtactics, Can you please look into it ?

specialtactics:
Hey @sudkumar sorry mate, I've been very busy, I will look at this shortly!

sudkumar:
Thank you. @specialtactics  I used the PR's commit till now.

I did not find any test case that I should update. So please let me know where/how I should write one if needed.

specialtactics:
@sudkumar I don't know why, but github says there's conflicts, so pull master and update please. I've had a look, and everything looks fine.

The laravel service provider actually doesn't even have this, which is odd - I will go and look into that separately now as well.

sudkumar:
Alright. I'll update.

sudkumar:
Hi @specialtactics . I swear I did not closed this. It says the commits have been merged!! I did not get it :/

specialtactics:
Hey @sudkumar  I think you've done something funny while pulling in master :stuck_out_tongue: 

Maybe just sync your repo and make a fresh PR ? Alternatively I could just make the change for you.

specialtactics:
Actually @sudkumar it seems like it was in fact already merged. Very weird github says closed instead of merged!

sudkumar:
I only took pull from dingo/api#master and pushed it back. No idea. =D

But it is merged and  that's what we need. Thank you. :)

specialtactics:
I added the same to the LaravelServiceProvider now:
https://github.com/dingo/api/pull/1666

Thanks for your contribution !

PR #1655, opened 2019-05-07
Allow https route declaration with 'https' array key and value

gazben:
Allow https route declaration with 'https' array key and value.
```
$router->group(['https' => true], function (Router $router) {
});
```

### DIFF
--- a/src/Routing/Route.php
+++ b/src/Routing/Route.php
@@ -511,7 +511,8 @@ class Route extends \Illuminate\Routing\Route
      */
     public function httpOnly()
     {
-        return in_array('http', $this->action, true);
+        return in_array('https', $this->action, true)
+            || (array_key_exists('https', $this->action) && $this->action['https']);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


gazben:
@specialtactics Can you merge this? Or do you think it will beak stuff?

specialtactics:
Can't imagine any issues, but I guess we will have to see! I ran tests on a couple of projects I have and no issues, so merging.

Derjuju:
Is this not a breaking change that prevent upgrading to Laravel/Lumen 6.x ?
#1689
Illuminate\Routing\Route return true/false on httpOnly if '**http**' is in action array
Dingo\Api\Routing\Route return true/false on httpOnly if '**https**' is in action array

Dingo\Api\Routing\Route extends Illuminate\Routing\Route

httpOnly, httpsOnly and secure methods shouldn't be overrided. They should keep the parent class logic.

Returning true with httpOnly when 'https' is set, is a non-sense... I think.

gazben:
@Derjuju  Oh my god, I messed this up. You are right. It should go to the httpsOnly/secure() function. I'll create the pull request

PR #1672, opened 2019-08-22
Fix escaping multibyte Unicode characters.

cryptour:
Encode multibyte Unicode characters literally (default is to escape as \uXXXX).
e.g. Chinese, Japanese.

Screen Shot 2019-08-22 at 8 35 37 PM

Screen Shot 2019-08-22 at 8 34 33 PM


specialtactics:
Thanks @cryptour

PR #1677, opened 2019-09-11
Get the original message of the Exception

jcardenascentral:


specialtactics:
Hi @jcardenascentral  - that looks good, however you will need to fix the style issues with the code, and have a look at the tests that are failing (looks a bit odd).

PR #1693, opened 2019-12-20
Add array route support

faytekin:
To have more readable codes, small changes have been done. In Route, Controller names are given as arrays.

Example
```php
use App\Api\V1\Controllers\LoginController;
$api->post('login', [LoginController::class, 'login']);
```

### DIFF
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -336,6 +336,7 @@ class Router
         } elseif ($action instanceof Closure) {
             $action = [$action];
         } elseif (is_array($action)) {
+            // For this sort of syntax $api->post('login', [LoginController::class, 'login']);
             if (is_string(Arr::first($action)) && class_exists(Arr::first($action)) && count($action) == 2) {
                 $action = implode('@', $action);
                 $action = ['uses' => $action, 'controller' => $action];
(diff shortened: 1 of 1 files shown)


taliptako:
@jasonlew @specialtactics Can you please merge this?

specialtactics:
Sorry, but I don't agree that this is more readable - it creates ambiguity because it is not clear what those array elements mean and what they relate to.

It is also not the way documented on the Laravel docs.

faytekin:
I think there has been a misunderstanding. You can set a route in this way at @specialtactics Laravel's own router. But this feature does not exist in Dingo API. I made it as an option. It does not break the current router. Anyone may use or just continue the current one.

src/Routing/Router.php
```diff
@@ -337,6 +337,11 @@ public function addRoute($methods, $uri, $action)
             $action = ['uses' => $action, 'controller' => $action];
         } elseif ($action instanceof Closure) {
             $action = [$action];
+        } elseif (is_array($action)) {
```
specialtactics: Can you just add a comment here,

```php
// For this sort of syntax $api->post('login', [LoginController::class, 'login']);
```

specialtactics (changes requested):

specialtactics:
Hey @faytekin that's fine in that case, I just requested one change which is to put a comment donoting what the code does.

Also there is a merge conflict with that file which you need to resolve.

specialtactics (approved):

specialtactics:
Merged, tagged

PR #1694, opened 2019-12-23
Add Route::patterns() while app is beeing bootstrapped and not after

christoph-kluge:
In my concrete scenario I was not able to separate my handlers based on the routes with different patterns. This PR allows us to handle it properly for the LaravelAdapter.

The following examples are working now properly:

```
$api->get('/entity/{uuid}', GetEntityByUuid::class)
$api->get('/entity/{slug}', GetEntityBySlug::class)
$api->get('/entity/{customIdentifier}', GetEntityByCustomIdentifier::class)
```

### DIFF
--- a/src/Routing/Adapter/Laravel.php
+++ b/src/Routing/Adapter/Laravel.php
@@ -2,12 +2,12 @@
 
 namespace Dingo\Api\Routing\Adapter;
 
+use Dingo\Api\Contract\Routing\Adapter;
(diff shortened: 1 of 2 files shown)


christoph-kluge:
referencing pull-request #1619 and issue #1232

specialtactics:
Hey @christoph-kluge 

Thanks for the PR, I did some local testing against projecting using UUIDs, and it seems to be working.

I'll tag as minor.

PR #1696, opened 2019-12-25
Refresh RouteCollection before dispatching to the route actions

### LINKED ISSUE #1695: route alias does not when used with name() function
| Q                 | A
| ----------------- | ---
| Bug?              | no|yes
| New Feature?      | no|yes
| Framework         | Laravel
| Framework version | 5.8.x
| Package version   | 2.4.0
| PHP version       | 7.3.x

#### Actual Behaviour

Given the following route definition:

```
/** @var Router $api */
$api = app(Router::class);
$api->version('v1', function (Router $api) {
    $api->get('entity/{uuid}', MyHandler::class)->name('myalias');
});
```

When using `route('myalias')` it will throw an exception `Route [myalias] not defined.`.

#### Expected Behaviour

Obviously it should not throw an exception

#### Steps to Reproduce

See actual behavior.

#### Possible Solutions

After a quick check I've realised that RouteCollection->add() does the name list lookup. The problem is that `->name()` is added afterwards to the route, so the route collection is not build up correctly. Refreshing the RouteCollection seems to fix the behavior. 

#### Additional note

It will still fail when we have an exception before the Adapter->dispatch() is hit and we'll use route()  during exception handling. Didn't found a better solution for now but it gives us an additional feature for now w/o breaking existing functionality.

christoph-kluge:
After merging RouteCollection we should ensure that we update the internal `actionList` and `nameList`. This allows us to use the internal lookups by name which is used by the `UrlGenerator->route()` member function aka `route()` helper function.

Resolves #1695

### DIFF
--- a/src/Routing/Adapter/Laravel.php
+++ b/src/Routing/Adapter/Laravel.php
@@ -2,12 +2,12 @@
 
 namespace Dingo\Api\Routing\Adapter;
 
+use Dingo\Api\Contract\Routing\Adapter;
+use Dingo\Api\Exception\UnknownVersionException;
 use Illuminate\Http\Request;
 use Illuminate\Routing\Route;
-use Illuminate\Routing\Router;
-use Dingo\Api\Contract\Routing\Adapter;
 use Illuminate\Routing\RouteCollection;
-use Dingo\Api\Exception\UnknownVersionException;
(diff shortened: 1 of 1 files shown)


specialtactics:
Nice @christoph-kluge  !

PR #1703, opened 2020-01-29
Fix http/https only detection

gazben:
This was a mistake that the httpsOnly logic was added to the http method. This is a fix for the breaking change in #1655


gazben:
Sorry about the last pull request somehow I didn't see that the logic is going to the wrong function.

specialtactics:
Thanks @gazben

PR #1714, opened 2020-03-08
Allow Laravel 7

stephane-monnot:


stephane-monnot:
How can I manage the declaration compatibility error ? 

Laravel 7 requires Throwable as argument type but Laravel 6 and 5 requires Exception.

Maybe create a new tag version only compatible with Laravel7 ?

PR #1715, opened 2020-03-09
[3.x] Laravel 7 Support

loonpwn:
This is a potentially breaking update.  With Laravel 7 there is new min requirements of PHP 7.2.5 (https://laravel.com/docs/7.x/upgrade#upgrade-7.0).

There is no way for dingo/api to work with PHP < 7 with implementing Throwable, not sure if that's an issue.

Note: `dingo/blueprint` has been removed as a core dependency. Ideally it wouldn't be a core package anymore and just exist as an optionally add-on. For now it can be re-added once there's a version with l7 support.


composer.json
```diff
@@ -13,24 +13,23 @@
         "email": "jason.lewis1991@gmail.com"
     }],
     "require": {
-        "php": "^7.1",
-        "dingo/blueprint": "^0.3",
```
lloricode: This is not support in laravel 7

lloricode:
Hi laravel/lumen 7 not install here, that why no error occurs.

`dingo/blueprint` need first to supprt laravel 7

lloricode:
you can check installed laravel version on your local, before running test

specialtactics:
Good work on this

I believe the blueprint package is needed for the docs generation, so we cannot drop that at this stage. Can you please put it back in, we will need to make it laravel 7 compatible also.

Also fix conflicts.

I created a new branch to work on laravel 7 compatibility, will get this merged into it.

specialtactics:
I tagged blueprint as v0.4.0 for laravel 7.
https://github.com/dingo/blueprint/releases/tag/v0.4.0

specialtactics:
Great, thanks !

PR #1716, opened 2020-03-09
[3.x] PHPUnit 8.5

loonpwn:
See this PR first: https://github.com/dingo/api/pull/1715

Laravel 7 has a min requirement of phpunit 8.5, so makes sense to update this to 8.5 as well.

While fixing the PHPUnit deprecations i've cleaned up the test files. Goal was to get the tests running but also to fix the issues stopping the IDE intellisense to work.
- Added a base test class to provide the `Mockery::close` and response resetting. The new phpunit uses typed void functions which required going through and updated all of the `tearDown()` and `setUp()` 
- Fixed `::class` invocations of classes which didn't exist for the namespaces.
- Cleaned up the imports
- Fixed PHPUnit's `@throwsException deprecation`


StringKe:
How to use your modified package temporarily?

lloricode:
clone, 
add this to composer.json
```json
  "repositories": [
    {
      "type": "vcs",
      "url": "<ssh|http>.git"
    }
  ],
```
then push on your own repo, the use it

specialtactics:
Great work !

PR #1724, opened 2020-03-19
Implement a proper array response function, which uses a transformer

specialtactics:
This will pass array responses through the normal transformation and wrap it inside the "data" key

### DIFF
--- a/src/Http/Response/Factory.php
+++ b/src/Http/Response/Factory.php
@@ -141,6 +141,32 @@ class Factory
     }
 
     /**
+     * Bind an arbitrary array to a transformer and start building a response
+     *
+     * @param array $array
+     * @param $transformer
+     * @param array $parameters
+     * @param Closure|null $after
+     * @return Response
+     */
+    public function array(array $array, $transformer, $parameters = [], Closure $after = null)
+    {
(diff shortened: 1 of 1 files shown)


RoyHP:
@specialtactics This change broke our API. We did not anticipate a backwards-breaking change like this in a SEMVER patch release: "Argument 1 passed to GovWorks\\Transformers\\Facility\\FacilityTransformer::transform() must be an instance of GovWorks\\Models\\Facility, instance of stdClass given, called in \/var\/www\/html\/vendor\/league\/fractal\/src\/Scope.php on line 407"

We had the following implementation. $facilities is a laravel collection of Facility models.
Screen Shot 2020-05-01 at 2 52 10 PM
Screen Shot 2020-05-01 at 2 52 23 PM
Screen Shot 2020-05-01 at 2 53 46 PM

specialtactics:
@RoyHP I do apologise that's happened for you, but it's not clear to me why you are using that function in that way.

It does not make sense to pass an array to a transformer which transforms a model.

You should either use response->item, or you should change your transformer to accept Stdclass.

RoyHP:
@specialtactics Yeah I'm not saying we're using it correctly, just that it has had this side effect. We will update the implementation to use item. Thank you.

PR #1742, opened 2020-08-28
Laravel Sanctum For Spa Provider

qazz92:
in config/api.php

` 'auth' => [
        'sanctum' => \Dingo\Api\Auth\Provider\SanctumSPA::class
    ],`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,6 @@
         "league/fractal": "^0.19"
     },
     "require-dev": {
-        "phpdocumentor/reflection-docblock": "3.3.2",
         "friendsofphp/php-cs-fixer": "~2",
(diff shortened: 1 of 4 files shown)


specialtactics:
Hey thanks @qazz92 - would you mind just also listing some install instructions? I see there is that middleware in api.php config, is there anything else if someone wants to use it?

pxdnbluesoul:
While this is a needed enhancement for the package I personally don't like some of the details of implementation, like commented code in the `config/api.php` file instead of adding it to the wiki as documentation. If this omitted that and the unnecessary change to the composer.json file and added the relevant info to the Wiki, it would be great.

PR #1758, opened 2020-10-19
Only load includes which correspond to actual relationships (fix #1096)

### LINKED ISSUE #1096: Including custom data in transformer
Hello there,

I have a repository that queries an API for data (rather than using a model) and returns either an object or a collection. I've been trying to include this data in a transformer for another model with limited success. Obviously I cannot us includes as this requires a model relationship, so for the moment I've hackishly grabbed my interface and checking whether a property exists on the model being transformed. My only issue is that the item is not rendering... Here's more "dodgy" code:

```PHP
public function transform(Domain $domain)
    {
        $output = [
            'id' => (int) $domain->id,
            'client_id' => (int) $domain->client_id,
            'url' => (string) $domain->url
        ];

        if ($id = $domain->dns_id) {
            $zone = app(ZoneRepositoryInterface::class);
            $output['dns'] = $this->item($zone->getById($id), new ZoneTransformer);
        }

        return $output;
    }
```

`$this->item($zone->getById($id), new ZoneTransformer)` appears to be the issue, it's calling the item successfully (can see object when debugging) but it's not outputting the `ZoneTransformer`, just a couple of `{}`.

Is there a better way of doing this? Any reason the item is not rendering?

Thanks

blmage:
In my use case, I have a `Language` model, whose display name can be translated into any other language:
```php
/**
 * @property int $id
 * @property string $code
 * @property-read string $self_display_name
 */
class Language extends RestfulModel
{
    protected $appends = [
        'self_display_name',
    ];

    /**
     * Returns the own displayable name of the language.
     *
     * @return string
     */
    public function getSelfDisplayNameAttribute(): string
    {
        return $this->getDisplayName($this->code);
    }

    /**
     * Returns a displayable name for the language, translated into another given language.
     *
     * @param string $languageCode
     * @return string
     */
    public function getDisplayName(string $languageCode): string
    {
        return Str::ucfirst(Str::lower(\Locale::getDisplayLanguage($this->code, $languageCode)));
    }
}
```
and a transformer that allows the client to specify the languages into which the display names should be translated:
```php
class LanguageTransformer extends BaseTransformer
{
    protected $availableIncludes = [
        'displayNames',
    ];

    /**
     * @param Language $language
     * @param ParamBag|null $params
     * @return ResourceAbstract
     */
    public function includeDisplayNames(Language $language, ParamBag $params = null)
    {
        if (null === $params) {
            return $this->null();
        }

        $displayNames = [];
        $languageCodes = array_filter(array_map('trim', (array) $params->get('languages')));

        foreach ($languageCodes as $languageCode) {
            $displayNames[$languageCode] = $language->getDisplayName($languageCode);
        }

        return $this->primitive($displayNames);
    }
}
```
But because  `displayNames` is not a relationship, queries fail whenever the corresponding include is requested.

### DIFF
--- a/src/Transformer/Adapter/Fractal.php
+++ b/src/Transformer/Adapter/Fractal.php
@@ -95,6 +95,19 @@ class Fractal implements Adapter
                 $eagerLoads = array_intersect($eagerLoads, $transformer->getAvailableIncludes());
             }
 
+            if (!empty($eagerLoads)) {
+                // Only load includes which correspond to actual relationships
+                $item = $response->first();
+
+                if ($item) {
+                    foreach ($eagerLoads as $key => $eagerLoad) {
+                        if (!method_exists($item, $eagerLoad)) {
+                            unset($eagerLoads[$key]);
+                        }
+                    }
+                }
+            }
+
             $response->load($eagerLoads);
         }
 
--- a/src/Transformer/Adapter/Fractal.php
+++ b/src/Transformer/Adapter/Fractal.php
@@ -95,13 +95,13 @@ class Fractal implements Adapter
                 $eagerLoads = array_intersect($eagerLoads, $transformer->getAvailableIncludes());
             }
 
-            if (!empty($eagerLoads)) {
+            if (! empty($eagerLoads)) {
                 // Only load includes which correspond to actual relationships
                 $item = $response->first();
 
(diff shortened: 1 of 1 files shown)


Milobut (changes requested):
Ji

PR #1762, opened 2020-11-30
More compatibility with Lumen

SpinyMan:
FormRequest affected


specialtactics:
Hey @SpinyMan thanks

I realise it may look more neat, but can you please not abstract the form request to a trait, I think that will cause problems either now or down the track. Basically keep both files separate, even if the contents are the same.

SpinyMan:
> I realise it may look more neat, but can you please not abstract the form request to a trait, I think that will cause problems either now or down the track. Basically keep both files separate, even if the contents are the same.

Hi,
Sorry, dude, but I disagree with you) The trait here is a good solution: if someday the `Lumen` and `Laravel` `FormRequest` became different you will always can create separated methods for each of them.

specialtactics:
@SpinyMan You can disagree all you like, that's irrelevant - instead, try to actually understand the point I'm making. Like I said I understand it's more neat, and it would be a good solution if this was bespoke code that nobody relied upon.

However as we are actually concerned with support, I am not merging in this code as it will obviously break implementations which check for traits of a class and don't take recursion like this into account. In the open source world you will find many examples of things not perfectly neat, done so because they want to support other packages and versions.

So you can either change the code, or I'm closing this PR.

SpinyMan:
OK, I just wanted to show the problem with FormRequest and Lumen... You can disapprove this PR and write your own that will resolve the issue. Ball is on your side, dude ;)

specialtactics:
I can't easily test the lumen aspect of this because I don't use it myself, and as I mentioned the travis integration is broken.

If there is demand for it, I'm sure somehow will do it, I'm very open to it, I just need the two providers' formrequest classes to be separate.

PR #1787, opened 2021-09-16
Feat: PHP8.0.8 threw an exception

Endy-c:
"uasort(): Returning bool from comparison function is deprecated, return an integer less than, equal to, or greater than zero"


specialtactics:
Hey @Endy-c if you are using this code for your project, I'm not so sure it's correct, you are supposed to use the spaceship operator.

In any case, this is fixed over here: https://github.com/api-ecosystem-for-laravel/dingo-api

PR #1788, opened 2021-09-28
Add #[ReturnTypeWillChange] to suppress errors and deprecation notices

rdeprera:
Return type of Dingo\Api\Routing\RouteCollection::count() isn't compatible with Countable::count()

On **PHP <var>7.4 >=</var>**
```console
dingo/api/src/Routing/RouteCollection.php on line 9

   Symfony\Component\ErrorHandler\Error\FatalError 

  During inheritance of Countable: Uncaught ErrorException: Return type of Dingo\Api\Routing\RouteCollection::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice
```
Same on getIterator()

### DIFF
--- a/src/Routing/RouteCollection.php
+++ b/src/Routing/RouteCollection.php
@@ -104,6 +104,8 @@ class RouteCollection implements Countable, IteratorAggregate
      *
      * @return \ArrayIterator
      */
+    
+    #[\ReturnTypeWillChange] 
     public function getIterator()
     {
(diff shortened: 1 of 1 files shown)


specialtactics:
Hey

I think you need to make it more clear where this error actually comes up, also can you please make this PR here:
https://github.com/api-ecosystem-for-laravel/dingo-api

PR #1793, opened 2022-01-12
Update default config :hammer:

richoandika:


specialtactics:
Not clear why this should be updated.

richoandika:
i think we should adding initial value, cause before I use without config its make error

specialtactics:
You need to add them to your env file, as per instructions to install. They can have an initial value of blank - which is actually the default, the fact that you get this error means you haven't added the necessary configs to your env file.

PR #1798, opened 2022-01-27
Add an abandoned note in the composer file, and update readme to poin…

specialtactics:
…t users to new repo

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,6 +38,7 @@
     "suggest": {
         "tymon/jwt-auth": "Protect your API with JSON Web Tokens."
     },
+    "abandoned": "api-ecosystem-for-laravel/dingo-api",
     "autoload": {
         "psr-4": {
             "Dingo\\Api\\": "src/"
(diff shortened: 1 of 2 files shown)


bytestream:
> Unfortunately this package cannot be maintained at this location anymore due to broken CI integrations, and travis-ci likely can't be used much longer either due to their change to paid plans.

Given you're a member of the Dingo org this seems a bit drastic. What's broken? 

`styleci` looks like it can be fixed via the config. If it can't then their support have been helpful for me in similar situations in the past. `travis-ci` can easily be replaced with GitHub actions.

specialtactics:
@bytestream It's not drastic, it's actually been the situation for many months - which has blocked any PRs from being merged, and I've tried to reach out to the author, and haven't had any response.

Styleci is not the problem, travis is the problem, and I can't fix it nor use github actions *as I'm not the owner of the repo.*

I think all these community repos should be maintained within an organisation to prevent this exact problem from happening.

